PR #2, opened 2022-05-09
DFA-691 Create Login Process

JohnWattsGDS:
Use sign-in form to sign user in and store token in session. Correct incorrect capitalisation of AuthenticationResult which meant the tokens were never stored.

Create basic authoriser middleware to check for presence of token and render login page if token missing.

Add authoriser middleware to router for list-services controller so that users must be logged in to view this screen.

Separate login with temporary password from display of change password view so it can be more easily re-used.


muthukswamy:
Needs fixing merge conflict

PR #9, opened 2022-05-18
DFA-727 DynamoDb Client

JohnWattsGDS:
Initial work on a reusable client for lambdas to make interacting with DynamoDb from lambdas very simple.

Create lambda to insert user
Create lambda to insert service
Remove example lambdas provided by SAM init
Remove example lambdas from template.yaml
Create an onboarding table type to help constrain what's sent to DDB
Use Axios to make lambda call and save user to DDB (this will be vastly improved using a facade/client to abstract away the details in the controller).


express/src/controllers/create-account.ts
```diff
@@ -168,7 +170,44 @@ export const submitMobileVerificationCode = async function (req: Request, res: R
...
+        console.log(response);
+
+        const URL = process.env.API_BASE_URL
+        const uuid = randomUUID();
+        const email = req.session.user?.UserAttributes?.filter((attribute: AttributeType) => attribute.Name === 'email')[0].Value;
+        const phone = req.session.mobileNumber;
+
+        let user = {
+                "pk": `user#${uuid}`,
+                "sk": `cognito_username#${req.session.user?.Username}`,
```
muthukswamy, line 185: Can we call it cognito_id instead? Username will be the email address so it might get confusing later.

lambda/dynamo-api/src/client/DynamoClient.ts
```diff
@@ -0,0 +1,31 @@
+import {DynamoDBClient, PutItemCommand, PutItemCommandOutput} from "@aws-sdk/client-dynamodb";
+import {marshall, unmarshall} from "@aws-sdk/util-dynamodb";
+import {OnboardingTableItem} from "../@Types/OnboardingTableItem";
+
+class DynamoClient {
+    private dynamodb: DynamoDBClient;
+    private readonly tableName: string;
+
+    constructor(tableName: string) {
+        this.dynamodb = new DynamoDBClient({ region: "eu-west-2" });
```
muthukswamy, line 10: Can we move the region to env file? It is less likely to change but just in case.

lambda/dynamo-api/src/handlers/put-user.ts
```diff
@@ -0,0 +1,32 @@
+import { APIGatewayProxyEvent, APIGatewayProxyResult } from 'aws-lambda';
```
muthukswamy, line 1: put-service and put-user files look mostly the same. Any way to reuse the code or better to keep them separate due to use in lambda?

express/src/controllers/create-account.ts
JohnWattsGDS: I went with username because that's the actual property used by Cognito but `cognito_id` works fine and I see your point.

lambda/dynamo-api/src/client/DynamoClient.ts
JohnWattsGDS: It would be an env var from the SAM template for a lambda but that can definitely be a configurable value.

JohnWattsGDS: In fact, AWS_REGION is available by default so we don't even have to configure it ...

muthukswamy (approved):
All good.

PR #10, opened 2022-05-18
Dfa 763 phone number sanitation middleware

tinokingstone:


express/src/middleware/mobileValidator.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+   })
+ 
+})
+ 
+describe('Checking that only numbers starting with +447[1-9] and 07[1-9] are accepted', () => {
+   it('An invalid +48 number is rejected.', async function () {
+       assert.equal(await ukNumberCheck(INVALID_NUMBER_STARTING_WITH_PLUS_48), false, `${INVALID_NUMBER_STARTING_WITH_PLUS_48} should not be allowed as it does not begin with +447`);
+   })
+ 
+   it('A valid +44 number is accepted.', async function () {
```
JohnWattsGDS, line 79: This needs clarifying - it says it tests that a valid +44 number is accepted, tests a number starting +4470 (which should not be accepted as valid - use a different example number for this method if necessary) and if the test were to fail it will report that "+4470whatever should be allowed as it does not begin with +447".  Except it does begin with +447.

JohnWattsGDS, line 79: I don't think `NUMBER_BEGINNING_WITH_PLUS_4470` is sensible test value to use here if we're testing that a valid +44 number is accepted and I don't think the failure message makes sense.

express/src/middleware/mobileValidator.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+       assert.equal(await ukNumberCheck(NUMBER_BEGINNING_WITH_PLUS_4470), true, `${NUMBER_BEGINNING_WITH_PLUS_4470} should be allowed as it does not begin with +447`);
+   })
+ 
+   it('An invalid 08 number is rejected.', async function () {
+       assert.equal(await ukNumberCheck(NUMBER_BEGINNING_WITH_08), false, `${NUMBER_BEGINNING_WITH_08} should not be allowed as it does not begin with +447`);
+   })
+ 
+   it('An valid 07 number is rejected.', async function () {
```
JohnWattsGDS, line 87: This needs looking at because we shouldn't reject valid 07 numbers.  The test doesn't use an 07 number in any case so this is very confusing..

express/src/middleware/mobileValidator.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+       assert.equal(await ukNumberCheck(INVALID_NUMBER_STARTING_WITH_PLUS_44), true, `${INVALID_NUMBER_STARTING_WITH_PLUS_44} should not be allowed as it does not begin with +447`);
+   })
+  
+})
+ 
+ 
+ 
+describe('Checking that only numbers starting with +447[1-9] and 07[1-9] are accepted', () => {
```
JohnWattsGDS, line 95: This describe line is copy pasted from above - if the tests inside it fit the description, we don't need another describe line.  If they don't then the line should be unique and describe what the tests are doing.

express/src/middleware/mobileValidator.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+ 
+   it('Checking that numbers which begin with 070 or +4470 are recognised . ', async function () {
+       assert.equal(await beginsWith070(NUMBER_BEGINNING_WITH_070), true, `${NUMBER_BEGINNING_WITH_070} should be recognised as a number beginning with 070`);
+       assert.equal(await beginsWith070(NUMBER_BEGINNING_WITH_PLUS_4470), true, `${NUMBER_BEGINNING_WITH_PLUS_4470}  should be recognised as a number beginning with 070`);
+   })
+ 
+   it('Checking that numbers which begin with 07 or +4470 are rejected. ', async function () {
```
JohnWattsGDS, line 102: 079 is valid but this says it should be rejected?

express/src/middleware/mobileValidator.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+       assert.equal(await beginsWith070(NUMBER_BEGINNING_WITH_070), true, `${NUMBER_BEGINNING_WITH_070} should not be rejected`);
```
JohnWattsGDS, line 103: It should be rejected though shouldn't it?  This is also the same code as above so it's not testing anything new.  If this was a test that ran the number through the entire validator then it would be a useful case.

express/src/middleware/mobileValidator.ts
```diff
@@ -1,16 +1,80 @@
...
+        errorResponse(mobileNumber, res , 'mobileNumber', 'We do not accept premium rate telephone numbers.');
+        return;
     }
-}
\ No newline at end of file
+
+    mobileNumber = convert07Numbers(mobileNumber);
+
+    if (!lengthCheck(mobileNumber)) {
+        errorResponse(mobileNumber, res , 'mobileNumber', 'Mobile number must be 10 digits long.');
```
JohnWattsGDS, line 29: It would be better to just say the number contains the incorrect number of digits instead of asking them to take off the 0 or +44 and then count the number of digits.

JohnWattsGDS (changes requested):
I've left some comments on some of the test cases which need clarifying

PR #13, opened 2022-06-05
Update Enter your email address page

BorisP-GDS:


express/src/middleware/emailValidator.ts
```diff
@@ -29,16 +29,15 @@ export async function emailValidator(req: Request, res: Response, next: NextFunc
 
 export function errorResponse(emailAddress: string, res: Response, key: string, message: string) {
     const errorMessages = new Map<string, string>();
-    const values = new Map<string, string>();
+    let values : object = {};
     errorMessages.set(key, message);
-    values.set('emailAddress', emailAddress);
+    values = {email: emailAddress};
```
JohnWattsGDS, line 34: We can probably do this as a one liner now - it was over two lines before because I wasn't sure how to initialise a Map with values; as it's now an object it can just be done in one go - just put `let` at the start of this line and delete line 32.

BorisP-GDS: Update done

JohnWattsGDS (dismissed):

PR #23, opened 2022-06-08
DFA-836: Resend mobile OTP

tinokingstone:
-mobile OTP is resent on click

-front end updates to match figma

-added cucumber tests

-clean up

### DIFF
--- a/express/features/mobile-verification-resend.feature
+++ b/express/features/mobile-verification-resend.feature
@@ -1,0 +1,22 @@
+Feature: A contact form to resend the verification code to mobile
+  Background:
+    Given that the user is on the `Check your phone` page
+
+  Scenario: User doesn’t receive their code
+    When they click on the "Not received a text message?" link 
+    Then they should be directed to the following page: "/create/resend-phone-code"
+
+  Scenario: The user wants to contact the service via slack
+    When they click on the "Not received a text message?" link 
+    And they click on the "Slack channel" link 
+    Then they should be directed to the following URL: "https://ukgovernmentdigital.slack.com/?redir=%2Farchives%2FC02AQUJ6WTC"
+
+  Scenario: The user wants to contact the service
+    When they click on the "Not received a text message?" link 
+    And they click on the "support form" link 
+    Then they should be directed to the following URL: "https://www.sign-in.service.gov.uk/support"
+
+  Scenario: The user wants the app to resend their code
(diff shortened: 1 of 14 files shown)


express/features/support/steps/mobile-verification-resend-steps.ts
```diff
@@ -0,0 +1,48 @@
+
+   
+import {Given, When} from "@cucumber/cucumber";
+
+Given('that the user is on the `Check your phone` page', async function () {
+  await this.goToPath('/create/get-email');
+  await this.page.type("#email", 'tessa.ting@foo-bar.gov.uk');
+  await this.page.waitForSelector('#submit');
+  await this.page.click('#submit')
+  await this.goToPath('/create/check-email');
```
JohnWattsGDS, line 10: We should generally only use `gotToPath` once in each feature.  The step code is code that automates browser interactions so we want to go to a page, click some buttons or links or fill in a form and submit it and test the final behaviour.  If we go to the page we expect a form submission to take us to, we're not really testing whether the form takes us where we expect when we submit it.

I'll guess that at later stages you might run into problems because some things that are expected to be in the session are missing.  You can incorporate the changes in the branch `BAU-stub_login` that I created this morning for Boris and change the email address used to register to `registering-successfully@gds.gov.uk` which should provide the necessary responses in the right place.

This set up code should just emulate what a user would do to get to the screen we're testing.

express/features/mobile-verification-resend.feature
```diff
@@ -0,0 +1,19 @@
+Feature: A contact form to resend the verification code to mobile
+  Background:
+    Given that the user is on the `Check your phone` page
+
+  Scenario: User doesn’t receive their code
+    When they select the "Not received a text message?" link 
+    Then they should be directed to the following URL: "http://localhost:3000/create/resend-phone-code"
```
JohnWattsGDS, line 7: These should use the `Then they should be directed to the following page: "/relative/path"` step from the shared-steps.  These will not work for test environments when we have them.  The URL form is for testing external websites rather than pages in our own application.  It's correct for Slack and the support form but not our own pages.

JohnWattsGDS (changes requested):
Please take a look at the comments and ask if you have any questions.

JohnWattsGDS (dismissed):
👍🏻

JohnWattsGDS (approved):

PR #45, opened 2022-06-24
DFA-943 list services controller route and view

JohnWattsGDS:
A user logging in will be shown a list of their services unless; they only have one in which case they'll be taken straight to the client screen; they don't have any in which case they'll be taken to the create service screen.
    
Modify User type - this is to get things working.  We should review everything and figure out where we do our unmarshalling - I expect the answer is in the DDB client.

Modify manage-accounts.ts controllers file.
* List services controller with basic conditional logic to determine whether services are listed, a new one must be created or clients are displayed.
* Extract (our) Client ID from generated client and append to redirect path as parameter.

Modify sign-in.ts - redirect to /account/list-services

Modify Lamdba Facade interface and implementations - add `listServices(userId: string, accessToken: string)`

Modify routes/manage-accounts.ts - add path parameter to `/service-client-details`

Modify / create super basic place-holder list services page.

Modify put-service-client.ts to return record that was inserted instead of response from DynamoDB

### DIFF
--- a/express/src/controllers/manage-account.ts
+++ b/express/src/controllers/manage-account.ts
@@ -55,7 +55,9 @@ export const processAddServiceForm = async function (req: Request, res: Response
 
     let newServiceData = JSON.parse(newServiceOutput?.data.output);
     const generatedClient = await lambdaFacade.generateClient(newServiceData.serviceId, service, newUser.email as string, req.session.authenticationResult?.AccessToken as string)
-    const clientId = JSON.parse(generatedClient.data.output.body).clientId;
+    const body = JSON.parse(generatedClient.data.output).body;
+    const clientId = JSON.parse(body).clientId;
+
     res.redirect(`/client-details/${clientId}`);
 }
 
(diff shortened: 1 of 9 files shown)


express/src/controllers/manage-account.ts
```diff
@@ -5,12 +5,30 @@ import {User} from "../../@types/User";
...
 
 export const listServices = async function(req: Request, res: Response) {
-    res.render('manage-account/list-services.njk');
+    const lambdaFacade: LambdaFacadeInterface = req.app.get("lambdaFacade");
+    if(req.session.selfServiceUser == undefined) {
+        res.render('there-is-a-problem.njk');
+        return;
+    }
+    const user: User = req.session.selfServiceUser;
+    const services = await lambdaFacade.listServices(user.pk.S as string, req.session.authenticationResult?.AccessToken as string);
```
muthukswamy, line 14: Is it possible for us to abstract the data that frontend gets so it does not have to get the value from `pk.S`.

muthukswamy: If not now then please include for later.

muthukswamy (approved):

express/src/controllers/manage-account.ts
JohnWattsGDS: Yeah - that was my quick and dirty, make it work solution.

We can use the aws utility unmarshall (or marshall I always forget which is which) - the question is whether we do it in the lambda that returns the DB result or in the facade that calls the lambda from within Express.  We should discuss it to choose which one we want to do.

PR #48, opened 2022-06-28
DFA-942 get service clients and enable their display

JohnWattsGDS:
Add listClients method to LambdaFacade
Update various client related views and controllers to deal with clientId path parameters.

There's a good chance we'll need to go `change-whatever/:serviceId/:clientId` to make things easier but this is enough for now to display things.

### DIFF
--- a/express/src/controllers/manage-account.ts
+++ b/express/src/controllers/manage-account.ts
@@ -56,7 +56,6 @@ export const processAddServiceForm = async function (req: Request, res: Response
     let newServiceData = JSON.parse(newServiceOutput?.data.output);
     const generatedClient = await lambdaFacade.generateClient(newServiceData.serviceId, service, newUser.email as string, req.session.authenticationResult?.AccessToken as string)
     const body = JSON.parse(generatedClient.data.output).body;
-    const clientId = JSON.parse(body).clientId;
-
-    res.redirect(`/client-details/${clientId}`);
+    const serviceId = JSON.parse(body).pk;
+    res.redirect(`/client-details/${serviceId.substring(8)}`);
 }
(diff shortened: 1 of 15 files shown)


express/src/controllers/manage-account.ts
```diff
@@ -56,7 +56,6 @@ export const processAddServiceForm = async function (req: Request, res: Response
     let newServiceData = JSON.parse(newServiceOutput?.data.output);
     const generatedClient = await lambdaFacade.generateClient(newServiceData.serviceId, service, newUser.email as string, req.session.authenticationResult?.AccessToken as string)
     const body = JSON.parse(generatedClient.data.output).body;
-    const clientId = JSON.parse(body).clientId;
-
-    res.redirect(`/client-details/${clientId}`);
+    const serviceId = JSON.parse(body).pk;
+    res.redirect(`/client-details/${serviceId.substring(8)}`);
```
CharlesIC, line 60: Why `.substring(8)`?

express/src/lib/lambda-facade/LambdaFacade.ts
```diff
@@ -63,5 +63,11 @@ class LambdaFacade implements LambdaFacadeInterface {
         const bareUserId = userId.substring(5);
         return await (await this.instance).get(`/Prod/get-services/${bareUserId}`);
     }
+
+    async listClients(serviceId: string, accessToken: string): Promise<AxiosResponse> {
+        console.log("ServiceId is: " + serviceId)
+        const bareServiceId = serviceId.startsWith("service#") ? serviceId.substring(8) : serviceId;
```
CharlesIC, line 69: What? Why?

express/src/controllers/manage-account.ts
JohnWattsGDS: When we get the service ID back from the DB it's prepended with `service#`.  The `#` breaks path parameters if we stick it straight in so we have to take it off.  When we tidy all of this up and refactor this should definitely made 'clean' and self-documenting.  I'm kind of racked with guilt for not at least using `const LENGTH_OF_SERVICE_PREFIX = 8;` or something.

CharlesIC: Or just split on `#`

PR #53, opened 2022-07-05
DFA-913 : update client name

tinokingstone:


express/src/routes/testing-routes.ts
```diff
@@ -22,14 +22,15 @@ router.post('/change-client-name/:serviceId/:selfServiceClientId/:clientId', asy
...
     try {
-        result = await facade.updateClient(req.params.serviceId, req.params.selfServiceClientId, req.params.clientId, {data: clientName}, req.session.authenticationResult?.AccessToken as string);
+        await facade.updateClient(req.params.serviceId, req.params.selfServiceClientId, req.params.clientId, {data: clientName}, req.session.authenticationResult?.AccessToken as string);
     } catch (error) {
         console.log(error)
         res.redirect('/there-is-a-problem');
         return;
     }
+    req.session.userDetailsUpdated = true;
+    req.session.updatedField = "Client name";
```
JohnWattsGDS, line 33: this should be lower case to read correctly when displayed

express/src/routes/manage-account.ts
```diff
@@ -18,10 +18,17 @@ router.get('/client-details/:serviceId', async (req, res) => {
     const selfServiceClientId = client.sk.substring("client#".length);
     const serviceId = req.params.serviceId;
     const authClientId = client.clientId;
+
+    let isUpated = false;
```
JohnWattsGDS, line 22: You could just use `req.session.userDetailsUpdated` in place of `isUpdated` on line 30 and do away with the variable and if-block.

However, you can also rely on the falsiness of `undefined` and replace `isUpdated` with `req.session.updatedField` instead of `req.session.userDetailsUpdated`.

And once you've done that, you can just use `updatedField` in the template's if-block instead of `userDetailsUpdated` and reduce the number of things to deal with in the session and variables that need to be passed to the template.

JohnWattsGDS (changes requested):
You can simplify this a bit - see the comments for how it can be done.

PR #56, opened 2022-07-07
DFA-915 : change public key 

tinokingstone:


JohnWattsGDS (dismissed):
This can be merged.

It would have been better to generate a public key from the value provided as validation but that can be added in another iteration.  At the moment we need to get all the fields done for the user testing tomorrow.

PR #57, opened 2022-07-07
DFA-1055: update redirect uri

tinokingstone:


express/src/middleware/urlValidator.ts
```diff
@@ -0,0 +1,10 @@
+import {NextFunction, Request, Response} from "express";
+
+type MiddlewareFunction<T, U, V> = (T: Request, U: Response, V: NextFunction) => void;
+const URL = require("url").URL;
```
JohnWattsGDS, line 4: As discussed this should be an import in keeping with the rest of the codebase.

express/src/middleware/urlValidator.ts
```diff
@@ -0,0 +1,10 @@
...
+export function isUriValid(uri: string) {
+    if (!uri.startsWith("http://") && !uri.startsWith("https://")) {
```
JohnWattsGDS, line 6: As also discussed, the URL class needs to be used and the protocol property compared, only allowing http for the special case that the hostname is localhost.

express/src/routes/testing-routes.ts
```diff
@@ -58,20 +57,31 @@ router.post('/change-redirect-URIs/:serviceId/:selfServiceClientId/:clientId', a
         return;
     }
 
-    // further error handling around a valid array of valid URIs
-
     const redirectUris = req.body.redirectURIs.split(" ");
+    for (let i = 0; i < redirectUris.length; i++) {
```
JohnWattsGDS, line 61: As discussed this should be middleware that's inserted into the route rather than inline code.  As middleware it can be re-used for the other URI client property.

JohnWattsGDS (changes requested):
Please see the comments which repeat what we discussed in Slack.

CharlesIC:
Replaced by #61

PR #59, opened 2022-07-07
DFA-1460: Create Account Profile page frontend

BorisP-GDS:


CharlesIC (commented):
This looks fine, just needs to be rebases on main to resolve conflcits and eliminate the merge commit from the PR

PR #60, opened 2022-07-07
DFA-914 Change user attributes

JohnWattsGDS:
Quick solution to save user attributes.  Has messy logic to deal with the delivery of ticked checkboxes and ensure openid is always an option.  We could / should consider displaying nicer versions of the attributes - currently we use the actual attribute names expcted by Auth.  This makes the logic simpler right now but looks a bit odd.

Alter offlineAccess to offline_access in template and route/controllers as that's the parameter the auth API expects.
Add code to call Auth API and update client configuration.
Switch to using join instead of an array.reduce operation.  Barely relevant to this story but I saw we should probably render attributes with a comma-space so made the change everywhere.

### DIFF
--- a/express/src/routes/manage-account.ts
+++ b/express/src/routes/manage-account.ts
@@ -27,24 +27,20 @@ router.get('/client-details/:serviceId', async (req, res) => {
         clientName: client.data,
         serviceName: client.service_name,
         clientId: client.clientId,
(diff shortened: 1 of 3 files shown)


BorisP-GDS (dismissed):
Looks good

CharlesIC (dismissed):

CharlesIC (approved):

PR #63, opened 2022-07-11
DFA-1458: Create frontend for 'Change your public key' page, updated version

BorisP-GDS:
Create frontend for 'Change your public key' page, updated version, including error messages display.

### DIFF
--- a/express/assets/css/app.scss
+++ b/express/assets/css/app.scss
@@ -187,3 +187,7 @@ code {
     }
   }
 }
+
+.current-public-key-container {
+  overflow-wrap: break-word;
+}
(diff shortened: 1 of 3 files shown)


CharlesIC (commented):
We should just replace the [existing view](https://github.com/alphagov/di-onboarding-self-service-experience/blob/e2170896720fae9c281127575e0e665d40826a11/express/src/views/dashboard/change-public-key.njk) `change-public-key.njk` instead of naming this `change-public-key-v2.njk`. We don't need to version views in this way, git provides sufficient versioning. The code might need to be changed in places where `change-public-key.njk` is used but theu shouldn't be too different.

BorisP-GDS:
If we replace the view now, the live page will start throwing errors because there is no back-end to work with the new file upload option which this view introduces. The scope of the story which this PR is for, is to create only the visuals.

CharlesIC (approved):

PR #65, opened 2022-07-12
DFA-1484 Create 'change your service name' page

BorisP-GDS:
Create 'change your service name' page
View and testing route for 'change your service name' page

### DIFF
--- a/express/src/routes/testing-routes.ts
+++ b/express/src/routes/testing-routes.ts
@@ -170,4 +170,22 @@ router.get('/account', (req, res) => {
     });
 });
 
+// Testing routes for Change your service name page
+router.get('/change-service-name', (req, res) => {
+    res.render("account/change-service-name.njk", {
+        serviceName: 'My juggling service'
+    });
+});
+
+router.post('/change-service-name', async (req, res) => {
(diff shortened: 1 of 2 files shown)


CharlesIC (dismissed):

express/src/routes/testing-routes.ts
```diff
@@ -170,4 +170,22 @@ router.get('/account', (req, res) => {
     });
 });
 
+// Testing routes for Change your service name page
+router.get('/change-service-name', (req, res) => {
+    res.render("account/change-service-name.njk", {
+        serviceName: 'My juggling service'
```
muthukswamy, line 176: Can the service name be taken from FB than hard coding a random name.

muthukswamy (changes requested):

express/src/routes/testing-routes.ts
BorisP-GDS: @muthukswamy  These are TESTING routes, the backend does not exist, the purpose of this routes is to visualize the forntend, so it can be tested by the BA, Design, and Content. When the backend is created the service name will be dynamic and coming from database or session. So for this purpose it does not really matter where does the service name come from.

PR #69, opened 2022-07-19
DFA-1494 Create 'Add your new password' page

BorisP-GDS:
Create 'Add your new password' page

### DIFF
--- a/express/assets/javascripts/show-password-two-fields.js
+++ b/express/assets/javascripts/show-password-two-fields.js
@@ -1,0 +1,85 @@
+window.GOVUK = window.GOVUK || {};
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+    function ShowPassword2($module) {
+        this.$module = $module
+        this.input = this.$module.querySelector('.gem-c-input-2')
+    }
+
+    ShowPassword2.prototype.init = function () {
+        this.$module.togglePassword = this.togglePassword.bind(this)
+        this.$module.revertToPasswordOnFormSubmit = this.revertToPasswordOnFormSubmit.bind(this)
+        this.input.classList.add('gem-c-input--with-password')
+
+        this.showPasswordText = this.$module.getAttribute('data-show-text')
+        this.hidePasswordText = this.$module.getAttribute('data-hide-text')
+        this.showPasswordFullText = this.$module.getAttribute('data-show-full-text')
+        this.hidePasswordFullText = this.$module.getAttribute('data-hide-full-text')
+        this.shownPassword = this.$module.getAttribute('data-announce-show')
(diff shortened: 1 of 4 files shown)


JohnWattsGDS (approved):

express/package.json
```diff
@@ -9,7 +9,7 @@
     "dev": "COGNITO_CLIENT=${COGNITO_CLIENT:-StubCognitoClient} LAMBDA_FACADE=${LAMBDA_FACADE:-StubLambdaFacade} nodemon --config nodemon.json -r 'dotenv/config' src/app.ts",
     "buildts": "tsc -p .",
     "buildsass": "node-sass --importer node_modules/node-sass-package-importer/dist/cli.js assets/css/app.scss dist/assets/app.css",
-    "buildjs": "uglifyjs --verbose node_modules/govuk-frontend/govuk/all.js -o dist/assets/govuk-all.js && uglifyjs --verbose assets/javascripts/cookies.js -o dist/assets/cookies.js && cp -v -r assets/javascripts/show-password.js ./dist/assets",
+    "buildjs": "uglifyjs --verbose node_modules/govuk-frontend/govuk/all.js -o dist/assets/govuk-all.js && uglifyjs --verbose assets/javascripts/cookies.js -o dist/assets/cookies.js && cp -v -r assets/javascripts/show-password.js ./dist/assets && cp -v -r assets/javascripts/show-password-two-fields.js ./dist/assets",
```
CharlesIC, line 12: Can we copy all the files from the `express/assets/javascripts` directory instead of having to add a copy command each time we add a new js file?

express/assets/javascripts/show-password-two-fields.js
```diff
@@ -0,0 +1,85 @@
+window.GOVUK = window.GOVUK || {};
```
CharlesIC, line 1: This file is identital to [show-password.js](https://github.com/alphagov/di-onboarding-self-service-experience/blob/b64ef29d6a7a0aa49579231f03381f46f43448ef/express/assets/javascripts/show-password.js). We should make the existing file/function reusable instead of duplicating big chunks of code.

express/package.json
BorisP-GDS: No at the moment we can not. Some of the files are optimized and some are written in a way that can not be optimized. That's why some of the files are optimized and some are copied, and we can not just move them all at once.

express/assets/javascripts/show-password-two-fields.js
BorisP-GDS: The file is not identical, it is similar. And it handles the second hide/show button. Some of the code might be re-usable - we can create  a tech debt for it. This was the quickest working solution to enable TWO show/hide password buttons on a same page.

PR #77, opened 2022-07-26
DFA-2025-Refactor header navigation to be dynamic - depending on user…

BorisP-GDS:
Refactor header navigation to be dynamic - depending on user being logged in or not.


JohnWattsGDS (approved):
Menus change when I log in and log out. 🚀

PR #79, opened 2022-07-27
DFA-2015 Align the content in the build to the content doc

BorisP-GDS:
Align the content in the build to the content doc [Sign in flow]

### DIFF
--- a/express/src/controllers/create-account.ts
+++ b/express/src/controllers/create-account.ts
@@ -61,27 +61,24 @@ export const checkEmailOtp = async function (req: Request, res: Response) {
         return;
     }
 
-    const any6CharacterLongPattern = /^.{6}$/;
-    const onlyNumbersPattern = /^[0-9]+$/;
+    const sixDigitsPattern = /^[0-9]{6}$/;
     const otpToTest = req.body['create-email-otp'];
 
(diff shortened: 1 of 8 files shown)


JohnWattsGDS (approved):

express/src/controllers/create-account.ts
```diff
@@ -61,27 +61,24 @@ export const checkEmailOtp = async function (req: Request, res: Response) {
...
-            value: value
+            errorMessages: errorMessages
         });
         return;
     }
 
-    if (!onlyNumbersPattern.test(otpToTest)) {
-        console.log("Otp code is not only numbers");
+    if (!sixDigitsPattern.test(otpToTest)) {
+        console.log("No otp code entered or number of symbols does not equal to 6");
```
JohnWattsGDS, line 79: Is this log message correct?  If there's no OTP, they'd have returned at line 75 wouldn't they?  Also, should we trim the OTP around line 65?  I seem to remember the field is bigger than 6 characters and users might paste some whitespace in maybe.

BorisP-GDS: Requested changes applied

JohnWattsGDS (approved):

PR #81, opened 2022-07-27
DFA-2030: Apply formatting to the acceptance tests

CharlesIC:
- Correctly construct URLs in feature tests 

Avoid failing tests due to double slashes caused by directly adding two strings. Use the URL class instead to handle URL construction.

---

- Remove duplicated code fragments by moving the repeated blocks into functions.
- Apply formatting
- Optimise imports
- Use one kind of quotes for strings (single quotes)

### DIFF
--- a/express/features/support/steps/shared-steps.ts
+++ b/express/features/support/steps/shared-steps.ts
@@ -1,6 +1,7 @@
-import {Given, When, Then} from "@cucumber/cucumber";
-import {strict as assert} from "assert";
-import {getLink, checkUrl} from './shared-functions';
+import { Given, Then, When } from '@cucumber/cucumber';
+import { strict as assert } from 'assert';
+import { Page } from "puppeteer";
+import { checkUrl, getLink } from './shared-functions';
 
 Given('that the user is on the {string} page', async function (route: string) {
     await this.goToPath(route);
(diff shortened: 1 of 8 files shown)


express/features/support/steps/shared-steps.ts
```diff
@@ -83,4 +71,17 @@ Then('the {string} link will point to the following URL: {string}', async functi
 Then('the {string} link will point to the following page: {string}', async function (linkText, expectedPage) {
     let link = await getLink(this.page, linkText);
     await checkUrl(this.page, link, expectedPage);
-});
\ No newline at end of file
+});
+
+async function checkErrorMessageDisplayedAboveElement(page: Page, errorLink: any, errorMessage: string, field: string) {
```
JohnWattsGDS, line 76: Given you're such a stickler for names - this should either be named to reflect that it actually checks both that the error message is displayed in the error summary panel and above the field with the error.

However, because of the other version that deals with radio-buttons which also checks in the error summary so that should probably be lifted into a separate function that can be re-used there also.

Whether you add an extra step for the summary in each test is down to you of course.

We might be able to change the template that renders the radios and their errors too so that the the href is the same but that can probably wait for another PR.

CharlesIC: Renamed method to match the step.

The href thing is for another PR....

JohnWattsGDS (approved):

PR #87, opened 2022-07-29
DFA-2030: Run unit and acceptance tests on PRs

CharlesIC:
-  Add workflows to run unit and acceptance tests on PRs
- Fix acceptance tests 

Correctly await when the submit button is clicked. Hard-coded waits cause the function to time out. Await the `waitForNavigation` function instead.

### DIFF
--- a/express/features/support/steps/shared-steps.ts
+++ b/express/features/support/steps/shared-steps.ts
@@ -1,33 +1,24 @@
 import { Given, Then, When } from '@cucumber/cucumber';
 import { strict as assert } from 'assert';
 import { Page } from "puppeteer";
-import { checkUrl, getLink } from './shared-functions';
+import { checkUrl, clickLink, clickSubmitButton, getButtonLink, getLink } from './shared-functions';
 
 Given('that the user is on the {string} page', async function (route: string) {
     await this.goToPath(route);
 });
 
 When('they click on the {string} link', async function (text: string) {
-    let links = await this.page.$x(`//a[contains(., '${text}')]`);
-    await Promise.all([
-        this.page.waitForNavigation({ timeout: 10000 }),
-        links[0].click()
-    ]);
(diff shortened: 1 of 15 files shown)


express/features/support/steps/shared-steps.ts
```diff
@@ -1,33 +1,24 @@
...
     await this.goToPath(route);
 });
 
 When('they click on the {string} link', async function (text: string) {
     let links = await this.page.$x(`//a[contains(., '${text}')]`);
-    await Promise.all([
-        this.page.waitForNavigation({ timeout: 10000 }),
-        links[0].click()
-    ]);
+    await clickLink(this.page, links[0]);
```
JohnWattsGDS, line 12: There's a `getLink` method in the shared functions which should have been used in the first place; as well as abstracting away the xpath stuff, it will blow up if there are more than one links on the page with the same text (which happened in the product pages and is why the extra check is in there).  I think I'd prefer it to using the `this.page.$x` stuff inline as it will generally protect us from unexpected and confusing failures in future if the content changes.

CharlesIC: OK I made a change to use the `getLink` method and created another for button-link, not sure if it can be refactored into one method but for now at least the `this.page.$x` stuff is one place and re-used.

JohnWattsGDS: Nice.  I think I originally went with explicitly calling out the fact it was a button link to make it easier at a later date to realise the 'button' is actually a link.  It may be that we could just treat it as any other link but it's probably right to make sure the thing we call a button link is that and doesn't become just a link at some point by accident. 🚀

JohnWattsGDS (dismissed):

BorisP-GDS (approved):

JohnWattsGDS (approved):

PR #89, opened 2022-08-01
DFA-2034: Fix cookie banner in non-production deployments

BorisP-GDS:


express/assets/javascripts/cookies.js
```diff
@@ -60,32 +61,44 @@ var cookieBanner = function () {
         });
     }
 
-    function getCookieDomain () {
-        const PROD_HOSTNAME = 'sign-in.service.gov.uk';
+    function isProdDomain() {
```
JohnWattsGDS, line 64: ❤️  I'm a big fan of this sort of thing. The less thinking I have to do, the better!

JohnWattsGDS (approved):

PR #91, opened 2022-08-03
BAU: Simply DynamoClient and add tests

CharlesIC:
- Simplify DynamoClient 
Refactor the helper methods to be more concise
Replace loops with maps
Add tests for the client - use jest fluent assertions
Use the table name from the environment in DynamoClient
Return promises to the runtime to AWS can await and handle failures if needed.

- Apply formatting and optimise imports 
Use single quotes

- Refactor StubCognitoClient

-  Enable debugging 
Use ts-jest, source map and a transform to enable debugging and correct stack traces when running tests.


backend/dynamo-api/src/handlers/get-service-clients.ts
```diff
@@ -1,29 +1,21 @@
...
-
 export const getServiceClientsHandler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
     const serviceId = event.pathParameters?.serviceId;
     if (serviceId === undefined) {
         return noUserIdResponse;
     }
 
-    let response = {statusCode: 200, body: JSON.stringify(serviceId)};
-    await client.getClients(serviceId)
+    return new DynamoClient()
```
JohnWattsGDS, line 10: It's considered best practice to instantiate as much as possible at the top level of the code so that it happens at init rather than invocation.  It may offend your sense of order and rightness but it was in the right place for a Lambda.

CharlesIC: It was so save lines of code but it makes sense to initialise things at init. Will change this back

JohnWattsGDS: It offends _my_ sense of order and rightness 😆 .  Even more so in Java when you start constructing outside the constructor but I guess it's not as bad here in TS land.

backend/dynamo-api/src/client/DynamoClient.ts
```diff
@@ -16,16 +16,20 @@ type AttributeNames = {[nameToken: string]: string}
...
         data: '#D'
     }
 
-    constructor(tableName: string) {
+    private readonly tableName: string;
+    private readonly dynamodb: DynamoDBClient;
+
+    constructor() {
+        this.tableName = process.env.TABLE!;
+        if (!this.tableName) {
```
JohnWattsGDS, line 28: 🆒

backend/dynamo-api/tests/client/DynamoClient.test.ts
```diff
@@ -2,6 +2,8 @@ import { strict as assert } from 'assert'
 import DynamoClient from "../../src/client/DynamoClient";
 
 describe('DynamoDB client', function () {
+    process.env.TABLE = "identities";
+
     const client = new DynamoClient("identities");
```
JohnWattsGDS, line 7: This appears to cause the test to fail.

PR #101, opened 2022-08-12
DFA-2064 implement 2FA for sign in

JohnWattsGDS:
New users have SMS MFA enabled when their phone is verified.

Necessary logic moved from login controller to sms otp controller.

Gratuitous addition of async error handling and a custom error class so we can start moving towards cleaner code and error handling.

Move some basic boiler plate logic out of controllers and into a middleware component.

Update a bunch of templates to work with new errorMessages and values objects instead of the previous maps.

Update some validation middleware to through appropriately crafted exceptions instead of rendering pages correctly.


express/src/controllers/create-account.ts
```diff
@@ -216,6 +216,8 @@ export const submitMobileVerificationCode = async function (req: Request, res: R
         const email = req.session.cognitoUser?.UserAttributes?.filter((attribute: AttributeType) => attribute.Name === 'email')[0].Value;
         const phone = req.session.enteredMobileNumber;
 
+        await cognitoClient.setMfaPreference(req.session.cognitoUser?.Username as string);
```
CharlesIC, line 215: What if `cognitoUser` is null?

express/src/controllers/sign-in.ts
```diff
@@ -26,16 +29,45 @@ export const showLoginOtpMobile = async function (req: Request, res: Response) {
...
 
-export const processLoginOtpMobile = async function (req: Request, res: Response) {
-    // TO DO add the functionality to process the login mobile otp
-    if (true) { // because OTP was correct and we've implemented that
+export const processLoginOtpMobile = async function(req: Request, res: Response) {
+
+    if(!req.session.cognitoUser?.Username || !req.body['sms-otp'] || !req.session.session) {
+        res.render('there-is-a-problem.njk');
+        return;
+    }
```
CharlesIC, line 44: We'll probably need to think of some logging for cases like this, or move param validation to a middleware?

express/src/controllers/sign-in.ts
```diff
@@ -26,16 +29,45 @@ export const showLoginOtpMobile = async function (req: Request, res: Response) {
...
+    try {
+        response = await cognitoClient.respondToMfaChallenge(req.session.cognitoUser?.Username as string, req.body['sms-otp'], req.session.session as string);
+    } catch (error) {
+        if(error instanceof CodeMismatchException) {
+            const errorMessages: Map<string, string> = new Map<string, string>();
+            errorMessages.set('smsOtp', 'The code you entered is not correct or has expired - enter it again or request a new code')
+            res.render('common/check-mobile.njk', {
+                mobileNumber: req.session.mobileNumber,
+                formActionUrl: "/sign-in-otp-mobile",
+                errorMessages: errorMessages
```
CharlesIC: This is a pattern that can be refactored into its own thing across the app

express/src/controllers/sign-in.ts
```diff
@@ -26,16 +29,45 @@ export const showLoginOtpMobile = async function (req: Request, res: Response) {
...
+            })
+        }
+    }
+    req.session.authenticationResult = response.AuthenticationResult;
+
+    const payload = (req.session.authenticationResult?.IdToken as string).split('.');
```
CharlesIC, line 67: What is `authenticationResult` is null? We would get a null pointer exception on the next line, wouldn't we?

express/package-lock.json
```diff
@@ -18,6 +18,7 @@
         "cookie-parser": "^1.4.6",
         "dotenv": "^10.0.0",
         "express": "^4.17.1",
+        "express-async-errors": "^3.1.1",
```
CharlesIC, line 21: Not sure we want to depend on this package. The last update was in 2018 and it has only 4 contributors. We should think of a different solution or copy and maintain the source code ourselves.

express/src/controllers/sign-in.ts
```diff
@@ -47,14 +48,18 @@ export const processLoginOtpMobile = async function(req: Request, res: Response)
...
             res.render('common/check-mobile.njk', {
                 mobileNumber: req.session.mobileNumber,
                 formActionUrl: "/sign-in-otp-mobile",
                 errorMessages: errorMessages
             })
+            return;
+        }
+        else {
+            throw error;
         }
```
CharlesIC, line 63: This perhaps can be handled by some middleware too to move out exception handling out of the controllers?

express/src/app.ts
```diff
@@ -66,11 +66,13 @@ app.get("/", function (req: Request, res: Response) {
...
-    console.log("Error handler");
-    console.error(err)
-    res.send('This should be the something went wrong page');
+app.use(function (err: Error | SelfServiceError, req: Request, res: Response, next: NextFunction) {
+    if(err instanceof SelfServiceError && err?.options) {
+        res.render(err.options.template, {values: err.options.values, errorMessages: err.options.errorMessages});
+    } else {
+        console.error(err);
+        res.render('./there-is-a-problem.njk');
+    }
```
CharlesIC, line 75: This is a start 👍

express/src/middleware/emailValidator.ts
```diff
@@ -2,46 +2,44 @@ import {NextFunction, Request, Response} from "express";
...
     return async (req: Request, res: Response, next: NextFunction) => {
         let emailAddress: string = req.body.emailAddress;
 
         emailAddress = emailAddress.trim();
 
         if (emailAddress === "" || emailAddress === undefined || emailAddress === null) {
-            await errorResponse(render, emailAddress, res, 'emailAddress', 'Enter your email address');
-            return;
+            throw new SelfServiceError("No email provided", {
+                ...template,
```
CharlesIC, line 17: Why not pass a string?

express/src/routes/create-account-or-sign-in.ts
```diff
@@ -16,13 +13,14 @@ import {
...
-import {passwordValidator} from "../middleware/passwordValidator";
+import {passwordValidator} from "../middleware/passwordValidator"
 import notOnCommonPasswordListValidator from "../middleware/notOnCommonPasswordListValidator";
+import "express-async-errors"
 
 const router = express.Router();
 
 router.get('/create/get-email', showGetEmailForm);
-router.post('/create/get-email', emailValidator('create-account/get-email.njk'), processGetEmailForm);
+router.post('/create/get-email', emailValidator({template: 'create-account/get-email.njk'}), processGetEmailForm);
```
CharlesIC, line 23: There's something weird about passing template names to utility classes. The error handling might be the wrong way round.

express/src/controllers/sign-in.ts
JohnWattsGDS: 100% moving stuff like this into middleware.  I just didn't want to change every method in the codebase in one PR 😆

CharlesIC (approved):

PR #110, opened 2022-08-30
DFA-2048: Set up repo for deployments

CharlesIC:
**Enable local debugging for SEE frontend** 

- Add the `--inspect` flag to nodemon to enable the debugger. This flags allows the IntelliJ, Chrome or any other Node debugger to attach itself to the running application or when debugging tests in the IDE.

- Enable sourcemap generation to allow debuggers to correctly track the application and set breakpoints, and to let IDEs debug tests.

- Add a switch to enable sourcemap generation as they will be needed for Fargate/Cloudwatch to log correctly.

- Stub return values in StubLambdaFacade and stub Cognito config to enable the full sign-in and registration journeys locally.

_Notes on local debugging:_
Recommend adding an `Attach to Node.js/Chrome` run configuration in IntelliJ to debug a locally running application. Or use the Node.js debugger built into Chrome (accessible from the dev tools window).

It's also possible to run and debug the app entirely in IntelliJ. This enables running the app easily by just pressing the button in the top toolbar.

 **Set up a monorepo** 

Share common config between projects:
  - Typescript settings
  - npm dev packages

Use npm workspaces:
  - Packages will be installed to the `node_modules` directory in the root of the project
  - Packages may be installed to `node_modules` directories in project spaces if there are version conflicts
  - Scripts can be run from the root of the repo
  - Common dev packages are specified in the root `packages.json` file. This ensures the same version is used across all projects and removes duplication
  - Delete unnecessary npm packages and trim down the list to the required minimum

Use Typescript project references:
  - Add one shared TS config at the root of the repo and extend in child projects as needed.

_References:_
 - npm workspaces
   https://docs.npmjs.com/cli/v7/using-npm/workspaces
 - Typescript project references
   https://www.typescriptlang.org/docs/handbook/project-references.html

**Don't hard code paths to npm modules**

npm modules are now installed in the root of the repo. Make the codebase more robust by using npm methods to derive the location of specific packages rather than hard coding a path.

- npm scripts: Use the `npm ls -p <package>` to get the path.

- Typescript code: Use the `require.resolve("<package>/package.json")` method to get the path to the root of the module.
See https://stackoverflow.com/questions/10111163/how-can-i-get-the-path-of-a-module-i-have-loaded-via-require-that-is-not-mine

- Sass: include the path to the npm module when compiling Sass and use paths relative from the module's root in Sass files.

**Set up common linting config** 

Add ESLint and Prettier config files in the root of the repo so the same rules are applied to all projects.
_Reference:_ https://typescript-eslint.io/docs/linting/typed-linting/monorepos/#one-tsconfigjson-per-package-and-an-optional-one-in-the-root

- Use the ESLint Typescript and Prettier plugin, and apply the recommended rules.

- Use the ESLint projects config to specify all projects that should have linting rules applied.
  The `tsconfig.eslint.json` file in the root of the repo specifies additional files to lint not included by other `tsconfig.json` files in the repo.

The rules are currently not enforced, though we may decide to do so in the future. For now, it is recommended to set a keyboard shortcut in the IDE of choice to apply the eslint/prettier suggestions or to configure the IDE to apply them automatically when saving files.

**Build and deploy the monorepo**

- Use the latest PaaS cleanup job to fix the delete deployment job.
- Modify the npm scripts to be aware of the CI environment.

When running in CI:
  - Minify the generated CSS and JS files
  - Don't generate Typescript source maps

---

- Enable source maps in Lambdas 

  Source maps are needed for Cloudwatch to display paths correctly in logs and stack traces
  _Reference:_ https://serverless.pub/aws-lambda-node-sourcemaps/


JohnWattsGDS (approved):
🚀

PR #117, opened 2022-09-07
DFA-2106 - Remove the change link from your account page

BorisP-GDS:


express/src/views/account/account.njk
```diff
@@ -41,15 +41,7 @@
...
-              actions: {
-              items: [
-                {
-                  href: '/change-email-address',
-                  text: "Change",
-                  visuallyHiddenText: "email address"
-                }
-              ]
-            }
+              actions: {}
```
CharlesIC, line 44: Do we need to specify an empty object or can we just not include it?

CharlesIC, line 44: The actions key can just be removed from the row.

BorisP-GDS: Fixed

CharlesIC (approved):

PR #122, opened 2022-09-12
DFA-2129 Abstract away backing services

JohnWattsGDS:
Continue abstracting backing services and shifting stuff out of the session

Rely on returned Authentication results for Cognito related stuff and get rid of 'CognitoUser' from session.  Create a class with static method to extract relevant claims from AuthenticationResults

Move other functionality into S4.


express/src/controllers/manage-account.ts
```diff
@@ -69,15 +76,18 @@ export const showChangePasswordForm = async function (req: Request, res: Respons
...
 };
 
-export const showAccount = async function (req: Request, res: Response) {
-    const payload = (req.session.authenticationResult?.IdToken as string).split("."); //duplcated code
-    const claims = Buffer.from(payload[1], "base64").toString("utf-8");
-    const cognitoId = JSON.parse(claims)["cognito:username"];
-    req.session.mobileNumber = JSON.parse(claims)["phone_number"];
+export const showAccount = async function (req: Request, res: Response, next: NextFunction) {
+    if (!req.session.authenticationResult) {
+        next(SelfServiceErrors.Redirect("sign-in.njk"));
```
CharlesIC, line 81: On giving it more thought, having `SelfServiceErrors.Redirect` and `SelfServiceErrors.Render` is not actually a good or useful pattern. Errors should indicate what went wrong, not what to do, and embedding handling instructions in them is abusing the system. The calling code or the handlers are responsible for knowing how to handle specific errors.

For cases such as these where something is missing from the session the handling should depend on whether it's something we expect to happen - in which case redirect or render an error page in the controller - or something we never expect to happen - in which case throw an error to be handled by the app's handlers (which would usually log and show the error page).

In this specific case, we probably don't expect the auth result to be missing so we could throw a custom exception type (say `AuthCredsMissing`) and the handler registered in `app.ts` would decide how to handle that (e.g. log and redirect to the sign in page, or the error page).

JohnWattsGDS: Changing this as discussed in later commits.

HarshVasudev (dismissed):

HarshVasudev (approved):

PR #124, opened 2022-09-13
BAU: Set up common test config

CharlesIC:
- Remove unnecessary async/await in tests
- Remove unnecessary param in phone validator
- Remove unnecessary escapes in phone validation regex
- Remove debug logging from phone validation
- Use `it.each` in phone validation tests instead of a for-each loop
---
-  Remove chai and only use jest assertions in tests
---
- Add common jest config 
  Add a base config to manage common settings for jest. Extend in derived configurations.
  Run jest tests in all projects from the root of the repo when running in GHA.
  Make paths to resource files relative so jest can locate them.
  Add a resources config class to locate the required resource files.
---
- Remove unnecessary `shortenKey` custom method from nunjucks. The `slice` method can be used directly in the template.


express/src/config/resources.ts
```diff
@@ -0,0 +1,10 @@
+import path from "path";
```
JohnWattsGDS, line 1: 🥇

JohnWattsGDS (approved):

PR #126, opened 2022-09-21
DFA 2123 implement dynamodb session storage

HarshVasudev:
- [npm: connect-dynamodb](https://www.npmjs.com/package/connect-dynamodb)  for session storage.
- The session table is created by a SAM Template
- The default in-memory implementation can be used if an env var is specified.
- A local implementation of DDB using docker-compose

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ gds aws <your-aws-account> -- sam deploy --parameter-overrides "AuthRegistration
 
 To install and run DynamoDB local for session storage.
 Download and install Docker desktop - https://www.docker.com/products/docker-desktop/.
-cd ./backend/dynamo-db/ docker-compose up - This will spin up the local dynamoDB and switch SESSION_LOCAL flag true in .env file to use local DDB storage
+cd ./backend/dynamo-db/ docker-compose up - This will spin up the local dynamoDB.
 ```
 
 ## Configure and run the application
--- a/README.md
+++ b/README.md
@@ -36,6 +36,10 @@ cd ../dynamo-api
 sam build
 gds aws <your-aws-account> -- sam deploy --parameter-overrides "AuthRegistrationBaseUrl=https://oidc.integration.account.gov.uk DynamoDbTableStackName=<the-name-you-chose>" --guided
 # Choose a decent stack name (name it after yourself for example) and accept the other defaults but tell it yes about lambdas that don't have authorisation
+
+To install and run DynamoDB local for session storage.
+Download and install Docker desktop - https://www.docker.com/products/docker-desktop/.
+cd ./backend/dynamo-db/ docker-compose up - This will spin up the local dynamoDB and switch SESSION_LOCAL flag true in .env file to use local DDB storage
 ```
 
 ## Configure and run the application
--- a/backend/dynamo-db/template.yaml
+++ b/backend/dynamo-db/template.yaml
@@ -6,6 +6,8 @@ Transform:
 
 Resources:
   Table:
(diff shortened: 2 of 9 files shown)


backend/dynamo-db/template.yaml
```diff
@@ -39,8 +41,8 @@ Resources:
           Projection:
             ProjectionType: ALL
 
-    # checkov:skip=CKV_AWS_28: "Ensure Dynamodb point in time recovery (backup) is enabled"
-    # checkov:skip=CKV_AWS_119: "Ensure DynamoDB Tables are encrypted using a KMS Customer Managed CMK"
+    # check:skip=CKV_AWS_28: "Ensure Dynamodb point in time recovery (backup) is enabled"
```
JohnWattsGDS, line 44: I think these will need to be 'inside' the resource; so above `Type` and at the same level of indentation as `Type`.

See how https://github.com/alphagov/di-onboarding-self-service-experience/blob/main/backend/dynamo-api/template.yaml#L19 has nested them.  Failing that you'll have to RTM to figure it out.

HarshVasudev: That's spot on @JohnWattsGDS. It is passing now.

README.md
```diff
@@ -36,6 +36,10 @@ cd ../dynamo-api
 sam build
 gds aws <your-aws-account> -- sam deploy --parameter-overrides "AuthRegistrationBaseUrl=https://oidc.integration.account.gov.uk DynamoDbTableStackName=<the-name-you-chose>" --guided
 # Choose a decent stack name (name it after yourself for example) and accept the other defaults but tell it yes about lambdas that don't have authorisation
+
+To install and run DynamoDB local for session storage.
+Download and install Docker desktop - https://www.docker.com/products/docker-desktop/.
+cd ./backend/dynamo-db/ docker-compose up - This will spin up the local dynamoDB and switch SESSION_LOCAL flag true in .env file to use local DDB storage
```
JohnWattsGDS, line 42: This is ambiguous I think - it reads like running `docker-compose up` will also set the flag to true but that's a separate step.  In either case I've suggested a change to the env var further down so check that before you change the README!

express/src/app.ts
```diff
@@ -13,9 +13,23 @@ import manageAccount from "./routes/manage-account";
...
+import session from "express-session";
+import connect_dynamodb from "connect-dynamodb";
+import AWS from "aws-sdk";
 
 const app = express();
 
+const dynamoDBSessionOptions = {
+    table: process.env.SESSIONS_TABLE,
+    AWSRegion: "eu-west-2",
+    client: process.env.SESSION_LOCAL === "true" ? new AWS.DynamoDB({endpoint: new AWS.Endpoint("http://localhost:8000")}) : ""
```
JohnWattsGDS, line 25: Can we wrap the code here in a function of some sort?  I think it would look neater and explain what we're doing slightly better.

Also, can you consider using an env var like `DYNAMO_DB_ENDPOINT`?  I'm pretty sure you can use a JavaScript-ism like
```
if ( process.env.DYNAMO_DB_ENDPOINT )
   { // return a client with the specified endpoint }
else
   { // return the default client }
```

Then it becomes optional with a sensible default and we have the option to specify a different endpoint (although I can't think why we'd want to right now).  Relying on `'true'` being correctly spelled etc. can be tricky.

JohnWattsGDS (changes requested):
I have left a couple of comments, if you could take a look that would be great.

express/src/app.ts
JohnWattsGDS: (Probably put the function in the lib folder somewhere and import it)

JohnWattsGDS (approved):
👍🏻

PR #131, opened 2022-09-26
DFA-2104 - Add acceptance test for Scenario: The user tries to change…

BorisP-GDS:
… their phone number but needs a new SMS code

### DIFF
--- a/express/features/your-account.feature
+++ b/express/features/your-account.feature
@@ -40,6 +40,43 @@ Feature:
     When they click on the link that points to "/create/resend-phone-code"
     Then they should see the text "Resend security code"
 
+  Scenario: The user tries to change their current password
 
+    Given the user has logged in and arrived on the your account page
+    When they click on the link that points to "/change-password"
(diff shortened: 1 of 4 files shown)


express/features/your-account.feature
```diff
@@ -25,6 +25,58 @@ Feature:
 
   Scenario: The user tries to change their phone number but needs a new SMS code
 
+    Given the user has logged in and arrived on the your account page
+    When they click on the link that points to "/change-phone-number"
+    Then they should be directed to the following page: "/change-phone-number"
+
+    Given that the user is on the "/change-phone-number" page
```
JohnWattsGDS, line 32: I think we can drop this step - we've checked they're on the right page above and this step effectively types the URL into the address bar which will work but might end up masking a bug in future.

BorisP-GDS: Done.

JohnWattsGDS (approved):
🚀

PR #134, opened 2022-09-28
DFA-2153 - Remove the Render and Redirect exceptions and replace them…

HarshVasudev:
… with the render/redirect methods.


JohnWattsGDS (changes requested):
`controllers/create-account.ts:22` there's a render error thrown - we should just `res.render` and `return`.

Once that's gone `lib/errors.ts` can be deleted and the subsequent compile errors can be cleaned up - there will be some work to do in the error handler at the end of `app.ts`

express/src/lib/errors.ts
```diff
@@ -2,35 +2,3 @@ export interface RenderOptions {
     values?: {[name: string]: string};
     errorMessages?: {[name: string]: string};
 }
-
```
JohnWattsGDS, line 5: It should be possible to delete this whole file.  You can inline the RenderOptions type in this only place it's used with something like:

```
export default function emailIsPresentInSession(
    template: string,
    renderOptions: {
        values?: {[name: string]: string};
        errorMessages?: {[name: string]: string};
    }
):
```
There's a good chance that middleware will be changed again at some point in the future so I think it's okay to put that type there for now.

JohnWattsGDS (changes requested):

express/src/lib/errors.ts
HarshVasudev: Sure. Done.

JohnWattsGDS (approved):
👍🏻

PR #136, opened 2022-10-03
DFA-2149 use only domain objects in express app

JohnWattsGDS:
Move Cognito and Lambda services into services directory with S4 (name to change when @CharlesIC decides what it should be).

Make some domain types and some small utils to convert from what DDB returns into domain types.

Use domain types when calling S4.

Use an Updates type in the DDB Client used by the lambdas.

Some linting and some phone number validation that was missed.


express/src/app.ts
```diff
@@ -60,6 +59,10 @@ declare module "express-session" {
 configureViews(app, path.join(__dirname, "../src/views"));
 
 app.use(setSignedInStatus);
+app.use(function (req: Request, res: Response, next: NextFunction) {
+    console.log(req.session);
+    next();
+});
```
CharlesIC, line 65: Debugging to remove?

CharlesIC (approved):

PR #138, opened 2022-10-05
DFA-2170 Update the client details page to include user name and pass…

BorisP-GDS:
Update the client details page to include user name and pasword

### DIFF
--- a/express/src/views/service-details/client-details.njk
+++ b/express/src/views/service-details/client-details.njk
@@ -102,6 +102,14 @@
       <p class="govuk-body">The page(s) you want your users to see after they sign out of your service.</p>
     </div>
   </details>
+  <div id="client-details-integration-credentials">
+    <h3 class="govuk-heading-s">Details to view the end user journey in integration</h3>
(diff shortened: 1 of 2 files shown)


express/src/views/service-details/client-details.njk
```diff
@@ -102,6 +102,14 @@
       <p class="govuk-body">The page(s) you want your users to see after they sign out of your service.</p>
     </div>
   </details>
+  <div id="client-details-integration-credentials">
+    <h3 class="govuk-heading-s">Details to view the end user journey in integration</h3>
+    <p class="govuk-body">During the integration process, you’ll be prompted to enter the following details to see the end user journey. This is so the public do not confuse it with the live journey.</p>
+    <p class="govuk-body">
+      <span class="govuk-!-font-weight-bold">Username</span>: integration-user<br>
+      <span class="govuk-!-font-weight-bold">Password</span>: winter2021
```
CharlesIC, line 110: Is this likely to ever change?

JohnWattsGDS (approved):
👍🏻 
If it ever changes we can make it a configuration value and it will be the same amount of work.  If it doesn't, we've saved ourselves some faffery.

HarshVasudev (approved):

PR #140, opened 2022-10-09
DFA-2014: Check if session has expired

HarshVasudev:
Redirecting to /sign-in-again page if session has expired.

### DIFF
--- a/express/src/controllers/sign-in.ts
+++ b/express/src/controllers/sign-in.ts
@@ -113,8 +113,7 @@ export const processEnterMobileForm = async function (req: Request, res: Respons
 
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore
-    const response = await s4.setPhoneNumber(req.session.emailAddress, mobileNumber);
-
+    await s4.setPhoneNumber(req.session.emailAddress, mobileNumber);
     // presumably that was fine so let's try to veerify the number
 
     const codeSent = await s4.sendMobileNumberVerificationCode(accessToken);
@@ -132,4 +131,8 @@ export const processEnterMobileForm = async function (req: Request, res: Respons
         textMessageNotReceivedUrl: "/resend-text-code",
         active: "sign-in"
     });
+};
+
+export const sessionTimeout = async function (req: Request, res: Response) {
+    res.render("session-timeout.njk");
 };
--- a/express/src/routes/sign-in.ts
+++ b/express/src/routes/sign-in.ts
@@ -8,7 +8,8 @@ import {
     showResendPhoneCodeForm,
     showSignInFormEmail,
     showSignInFormPassword,
-    signOut
+    signOut,
+    sessionTimeout
 } from "../controllers/sign-in";
 import emailIsPresentInSession from "../middleware/emailIsPresentInSession/emailIsPresentInSession";
 import {emailValidator} from "../middleware/emailValidator";
@@ -19,6 +20,7 @@ import {processLoginOtpMobile} from "../middleware/sign-in-middleware";
 const router = express.Router();
 
 router.get("/sign-in", showSignInFormEmail);
+router.get("/session-timeout", sessionTimeout);
 router.post("/sign-in", emailValidator("sign-in.njk"), processEmailAddress);
 router.get(
     "/sign-in-password",
--- a/express/src/middleware/authoriser.ts
+++ b/express/src/middleware/authoriser.ts
@@ -3,7 +3,8 @@ import SelfServiceServicesService from "../services/self-service-services-servic
 
 export async function checkAuthorisation(req: Request, res: Response, next: NextFunction) {
     if (!req.session.authenticationResult?.AccessToken) {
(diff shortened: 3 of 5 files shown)


express/src/middleware/sessionValidator.ts
```diff
@@ -0,0 +1,10 @@
+import {NextFunction, Request, Response} from "express";
+
+export default function sessionValidator(req: Request, res: Response, next: NextFunction) {
+    if (req.session.user || req.path === "/" || req.path === "/sign-in" || req.path === "/create/get-email") {
```
CharlesIC: We can't keep adding URLs here, this is not a sustainable approach. Instead only add this check to pages that require a session.

express/src/controllers/sign-in.ts
```diff
@@ -133,3 +133,7 @@ export const processEnterMobileForm = async function (req: Request, res: Respons
         active: "sign-in"
     });
 };
+
+export const signInAgain = async function (req: Request, res: Response) {
+    res.render("you-need-to-sign-in-again.njk", {});
```
CharlesIC, line 138:
```suggestion
    res.render("you-need-to-sign-in-again.njk");
``

We don't need to pass empty options, just omit the param.

File: express/src/views/you-need-to-sign-in-again.njk [hunk 3]
```diff
@@ -0,0 +1,12 @@
...
+{% from "govuk/components/input/macro.njk" import govukInput %}
+
+{% set active = "sign-in" %}
+
+{% block pageTitle %}Session Expired.{% endblock %}
+
+{% block mainContent %}
+  <h1 class="govuk-heading-l">Your session has expired, You need to sign in again</h1>
+  <p class="govuk-body"><a class="govuk-link" href="/sign-in">Sign-in</a>.</p>
+{% endblock %}
```
CharlesIC, line 12: This content doesn't seem correct, can you please check against the content designer's copy.

express/src/routes/sign-in.ts
```diff
@@ -19,6 +20,7 @@ import {processLoginOtpMobile} from "../middleware/sign-in-middleware";
 const router = express.Router();
 
 router.get("/sign-in", showSignInFormEmail);
+router.get("/you-need-to-sign-in-again", signInAgain);
```
CharlesIC, line 23: I believe the URL for the timeout page should be `/session-timeout`. Please check against the design spec.

express/src/middleware/sessionValidator.ts
HarshVasudev: Yes. I have removed this file in the next commit. So probably this does not need to be addressed.

express/src/controllers/sign-in.ts
HarshVasudev: Done.

express/src/routes/sign-in.ts
HarshVasudev: Done.

express/src/views/you-need-to-sign-in-again.njk
HarshVasudev: Done.

express/src/controllers/sign-in.ts
```diff
@@ -133,3 +133,7 @@ export const processEnterMobileForm = async function (req: Request, res: Respons
         active: "sign-in"
     });
 };
+
+export const signInAgain = async function (req: Request, res: Response) {
+    res.render("you-need-to-sign-in-again.njk");
+};
```
CharlesIC, line 139:
```suggestion
export const sessionTimeout = async function (req: Request, res: Response) {
    res.render("session-timeout.njk");
};
```

Please use the existing `session-timeout` template and delete the duplicate template.

express/src/views/you-need-to-sign-in-again.njk
```diff
@@ -0,0 +1,13 @@
...
+
+{% set active = "sign-in" %}
+
+{% block pageTitle %}Session timeout - GOV.UK Sign In{% endblock %}
+
+{% block mainContent %}
+  <h1 class="govuk-heading-l">You need to sign in again</h1>
+  <p class="govuk-body">To keep your account secure, we’ve signed you out because you did not do anything for 60 minutes.</p>
+  <p class="govuk-body"><a class="govuk-link" href="/sign-in">Sign in again</a>.</p>
+{% endblock %}
```
CharlesIC, line 13: Duplicate view. This can be removed.

CharlesIC (commented):
A small clean-up suggested

express/src/controllers/sign-in.ts
```diff
@@ -113,8 +113,7 @@ export const processEnterMobileForm = async function (req: Request, res: Respons
 
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore
```
CharlesIC, line 115: I think we can remove these ignore comments now?

CharlesIC (commented):
Looks good now 👍

CharlesIC (approved):

PR #145, opened 2022-10-12
DFA-2036 Create 'An account already exist' page

BorisP-GDS:


express/assets/css/app.scss
```diff
@@ -178,3 +178,7 @@ code {
   border-left: 5px solid $govuk-link-colour;
   padding: 24px 24px 0 24px;
 }
+
+main.main-account-exists {
+  padding-top: 15px;
+}
```
CharlesIC, line 184: We shouldn't need custom padding for this page. We should stick to the design system's padding unless we have a custom design (which we don't for this page).

HarshVasudev (approved):
Looks okay

express/src/views/create-account/account-exists.njk
```diff
@@ -0,0 +1,53 @@
+{% extends "../layout.njk" %}
+{% from "../macros/back-link.njk" import backLink %}
+{% from "../macros/error-summary.njk" import errorSummary %}
+{% from "govuk/components/input/macro.njk" import govukInput %}
+
+{% set backLink = backLink("/create/get-email") %}
+{% set pageTitle = "An account already exists" %}
+{% set mainClasses = "main-account-exists" %}
```
CharlesIC, line 8: We don't need to set custom padding

PR #151, opened 2022-10-18
DFA-2048: ECR repo template and SAM scripts

CharlesIC:
- Add frontend Dockerfile
- Add template for GHA role
- Add templatre for ECR repo
- Add scripts to apply the templates
- Use latest actions

### DIFF
--- a/ci/apply-sam-template.sh
+++ b/ci/apply-sam-template.sh
@@ -1,0 +1,66 @@
+#!/usr/bin/env bash
+set -e
+
+OPTION_REGEX="^--?.*"
+
+while [[ -n $1 ]]; do
+  case $1 in
+  -t | --template)
+    shift
+    TEMPLATE=$1
+    ;;
+  -s | --stack-name)
+    shift
+    STACK_NAME=$1
+    ;;
+  -p | --prefix)
+    shift
+    PREFIX=$1
+    ;;
+  -g | --tags)
+    TAGS=()
+    while [[ -n $2 ]] && ! [[ $2 =~ $OPTION_REGEX ]]; do
+      shift
+      TAGS+=("$1")
+    done
+    ;;
+  -o | --params)
+    PARAMS=()
+    while [[ -n $2 ]] && ! [[ $2 =~ $OPTION_REGEX ]]; do
+      shift
+      PARAMS+=("$1")
+    done
+    ;;
+  *)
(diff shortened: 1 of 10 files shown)


ci/sse-infrastructure.yml
```diff
@@ -0,0 +1,39 @@
...
+      ImageScanningConfiguration:
+        ScanOnPush: true
+      EncryptionConfiguration:
+        EncryptionType: KMS
+      Tags:
+        - Key: "application"
+          Value: "di-self-service"
+      LifecyclePolicy:
+        RegistryId: !Ref AWS::AccountId
+        LifecyclePolicyText: '{
```
JohnWattsGDS, line 19: So this deletes images older than five days?  Will this cause us any problems - I seem to remember hub key rotation crapping out at some point because images were deleted.  It makes perfect sense for previews but is it what we want for our production image?

ci/apply-sam-template.sh
```diff
@@ -0,0 +1,64 @@
...
+[[ ${#PARAMS[@]} -gt 0 ]] && echo "  with params ${PARAMS[*]}"
+
+sam deploy \
+  --stack-name "$STACK_NAME" \
+  --template-file "$TEMPLATE" \
+  --confirm-changeset \
+  --disable-rollback \
+  --resolve-s3 \
+  --s3-prefix "$PREFIX" \
+  --no-fail-on-empty-changeset \
```
JohnWattsGDS, line 61: You've done this before haven't you? 😆

ci/sse-infrastructure.yml
CharlesIC: This is only used for previews, the process for prod will be different. And if the image gets deleted and we need it again, the pipeline will just build and push it again.

ci/apply-sam-template.sh
CharlesIC: Too many times 😆

ci/sse-infrastructure.yml
CharlesIC: Think of it as a deplouyment artifact - once the app is deployed, we can trash it. Ideally we'd keep one image per branch until the deployment for the branch is deleted but this is probably too much effort that it's worth. The image can be easily re-built if needed instead.

JohnWattsGDS (approved):
🚀

PR #154, opened 2022-10-24
DFA-1023-update-service-name

HarshVasudev:
- Change service name and render client-details with success banner


JohnWattsGDS (approved):
Not sure how I feel about using hidden form params since we've used URL stuff elsewhere in the application but this seems to do what we need to get a product out. @CharlesIC can decide if it needs iterating to use path params at a later date. Good to go for now.

PR #155, opened 2022-10-27
DFA-1400 Update SASS Engine - Migrate from node-sass to sass

BorisP-GDS:


express/package.json
```diff
@@ -7,7 +7,7 @@
     "dev": "COGNITO_CLIENT=${COGNITO_CLIENT:-StubCognitoClient} LAMBDA_FACADE=${LAMBDA_FACADE:-StubLambdaFacade} nodemon --config nodemon.json --inspect -r 'dotenv/config' src/app.ts",
     "minifyjs": "for script in dist/assets/*.js; do uglifyjs $script -cmo $script; done",
     "buildts": "tsc $(${CI:-false} && echo --sourceMap false)",
-    "buildsass": "node-sass --output-style compressed --importer $(npm ls -p node-sass-package-importer)/dist/cli.js --include-path $(npm ls -p govuk-frontend)/.. assets/css/app.scss dist/assets/app.css $(${CI:-false} || echo --source-map true --source-map-contents)",
+    "buildsass": "sass --load-path=$(npm ls -p govuk-frontend)/.. --style compressed assets/css/app.scss dist/assets/app.css $(${CI:-false} && echo --no-source-map) $(${CI:-false} || echo --embed-sources)",
```
CharlesIC, line 10:
```suggestion
    "buildsass": "sass --load-path=$(npm ls -p govuk-frontend)/.. --style compressed assets/css/app.scss dist/assets/app.css $(${CI:-false} && echo --no-source-map || echo --embed-sources)",
```

The logic can be combined into a single statement

CharlesIC (commented):
I've run `npm run buildsass` on this branch and the new engine produces 20+ warnings.
Can you address them in this PR?

Mainly deprecation warnings about doing divsion outside of the calc function

express/package.json
BorisP-GDS: Warnings are related to gov uk design system core, so these are to be addressed by gov uk design system team, we can turn off deprecation warnings if needed.

BorisP-GDS: SASS module generates the maps by default, and if we do not want them to be generated - as is our case in CI, we need to echo  --no-source-map . On the other hand when we generate the maps for local debugging we do not need `--no-source-map` but we need `--embed-sources`.

Could you, please,  advise how to combine the logic above in a single statement ?

CharlesIC: Yes, the logic is the same, we can just merge the two checks into one, it works the same as your original version.

CharlesIC:
Fix for the linting error in alphagov/di-github-actions#26

CharlesIC:
Confirmed the deprecation warnings are related to the `govuk-frontend` package and we're on latest version, so we can ignore them for this PR.

express/package.json
BorisP-GDS: Updated, commits squashed.

CharlesIC (approved):

PR #171, opened 2022-11-21
DFA 2037 send email code

HarshVasudev:
- Custom message trigger for forgot password - Password reset flow.
- ES lint changes - added more safer type such as never, unknown in place of any - As suggested.
- Acceptance tests of the forgot password flow.

### DIFF
--- a/express/src/services/cognito/CognitoClient.ts
+++ b/express/src/services/cognito/CognitoClient.ts
@@ -16,7 +16,9 @@ import {
     RespondToAuthChallengeCommand,
     RespondToAuthChallengeCommandOutput,
     VerifyUserAttributeCommand,
-    VerifyUserAttributeCommandOutput
+    VerifyUserAttributeCommandOutput,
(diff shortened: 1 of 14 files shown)


express/src/services/cognito/CognitoClient.ts
```diff
@@ -122,6 +124,16 @@ If the app is being deployed to PaaS then you may have to update manifest.yaml o
         return await this.cognitoClient.send(command);
     }
 
+    async forgotPassword(email: string): Promise<ForgotPasswordCommandOutput> {
+        console.debug(`sending forgotPassword code to this address : ${{email}}`);
```
JohnWattsGDS, line 128: I don't think we want to log users' email addresses.

HarshVasudev: Removed.

JohnWattsGDS (approved):

BorisP-GDS (approved):

PR #174, opened 2022-11-25
DFA-2038 - resending the forgot password email when resend-email butt…

HarshVasudev:
- forgot password email is resend when resend email button is clicked.
- Added acceptance tests when resending of the password is requested
- Updated the html component to have a button id as "submit "so feature step can identify it in the DOM with a id

### DIFF
--- a/express/features/support/steps/shared-functions.ts
+++ b/express/features/support/steps/shared-functions.ts
@@ -18,20 +18,26 @@ export async function getButtonLink(page: Page, linkText: string): Promise<Eleme
     return getSingleLink(links, linkText);
 }
 
-export async function clickLink(page: Page, link: ElementHandle, timeout = DEFAULT_TIMEOUT): Promise<any> {
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+export async function clickLink(page: Page, link: ElementHandle, timeout = DEFAULT_TIMEOUT): Promise<unknown> {
     await Promise.all([page.waitForNavigation({timeout: timeout}), link.click()]);
 }
 
-export async function clickSubmitButton(page: Page, timeout = DEFAULT_TIMEOUT): Promise<any> {
-    await clickButtonWithId(page, "submit", timeout);
+export async function clickSubmitButton(page: Page, timeout = DEFAULT_TIMEOUT): Promise<unknown> {
+    return await clickButtonWithId(page, "submit", timeout);
 }
 
-export async function clickButtonWithId(page: Page, id: string, timeout = DEFAULT_TIMEOUT): Promise<any> {
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+export async function clickButtonWithId(page: Page, id: string, timeout = DEFAULT_TIMEOUT): Promise<never> {
     await Promise.all([page.waitForNavigation({timeout: timeout}), page.click(`#${id}`)]);
 }
 
 export async function checkUrl(page: Page, link: ElementHandle, expectedUrl: string): Promise<void> {
-    assert.equal(await page.evaluate((anchor: {getAttribute: (arg0: string) => any}) => anchor.getAttribute("href"), link), expectedUrl);
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    assert.equal(await page.evaluate((anchor: {getAttribute: (arg0: string) => never}) => anchor.getAttribute("href"), link), expectedUrl);
 }
 
 function getSingleLink(links: ElementHandle[], match: string): ElementHandle {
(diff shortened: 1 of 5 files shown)


express/src/controllers/sign-in.ts
```diff
@@ -116,6 +116,16 @@ export const forgotPasswordForm = async function (req: Request, res: Response) {
 };
 
 export const checkEmailPasswordReset = async function (req: Request, res: Response) {
+    await forgotPassword(req, res);
+    res.render("check-email-password-reset.njk");
+};
+
+export const resendForgotPassword = async function (req: Request, res: Response) {
+    await forgotPassword(req, res);
```
JohnWattsGDS, line 124: Does this behave properly if the password is an empty string?  I feel like this will generate one of the can't render twice type errors.

CharlesIC: @JohnWattsGDS Shouldn't the validation of email be done via middleware in the route?

express/src/routes/sign-in.ts
```diff
@@ -52,5 +53,6 @@ router.get("/no-account", (req, res) => {
 
 router.get("/forgot-password", forgotPasswordForm);
 router.get("/check-email-password-reset", checkEmailPasswordReset);
+router.post("/check-email-password-reset", resendForgotPassword);
```
CharlesIC, line 56: I believe `emailIsPresentInSession ` should be used for these routes (check with @JohnWattsGDS )

express/src/controllers/sign-in.ts
JohnWattsGDS: This would solve the problem. `middleware/validators/emailIsPresentInSession` can be inserted into the middleware chaing.

HarshVasudev: Done.

express/src/routes/sign-in.ts
HarshVasudev: Done.

express/src/controllers/sign-in.ts
```diff
@@ -129,3 +125,14 @@ export const checkEmailPasswordReset = async function (req: Request, res: Respon
...
 };
+
+export const resendForgotPassword = async function (req: Request, res: Response) {
+    const emailAddress = req.session.emailAddress;
+    try {
+        const s4: SelfServiceServicesService = await req.app.get("backing-service");
+        await s4.forgotPassword(emailAddress as string);
+    } catch (error) {
+        console.error("ERROR CALLING COGNITO - FORGOT PASSWORD WITH EMAIL", error);
+    }
```
CharlesIC, line 136: This can be refactored into a method now

express/src/routes/sign-in.ts
```diff
@@ -50,7 +51,20 @@ router.get("/no-account", (req, res) => {
     res.render("no-account-found.njk");
 });
 
-router.get("/forgot-password", forgotPasswordForm);
-router.get("/check-email-password-reset", checkEmailPasswordReset);
+router.get(
+    "/forgot-password",
+    emailIsPresentInSession("sign-in.njk", {errorMessages: {emailAddress: "Enter your email address"}}),
```
CharlesIC, line 56: We'll need to fix the repetition of error messages but that's a story for another time

CharlesIC (commented):
Please also squash the commits

express/src/controllers/sign-in.ts
HarshVasudev: Done.

express/features/support/steps/shared-functions.ts
```diff
@@ -18,28 +18,36 @@ export async function getButtonLink(page: Page, linkText: string): Promise<Eleme
     return getSingleLink(links, linkText);
 }
 
-export async function clickLink(page: Page, link: ElementHandle, timeout = DEFAULT_TIMEOUT): Promise<any> {
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+export async function clickLink(page: Page, link: ElementHandle, timeout = DEFAULT_TIMEOUT): Promise<unknown> {
```
CharlesIC, line 23: Which warning/error is being ignored here?

express/src/controllers/sign-in.ts
```diff
@@ -116,16 +116,21 @@ export const forgotPasswordForm = async function (req: Request, res: Response) {
...
+    await forgotPasswordCognitoCall(req);
+    res.render("check-email-password-reset.njk");
+};
+
+export const resendForgotPassword = async function (req: Request, res: Response) {
+    await forgotPasswordCognitoCall(req);
+    res.redirect("/create-new-password");
+};
+
+export const forgotPasswordCognitoCall = async function (req: Request) {
```
CharlesIC, line 128:
```suggestion
const forgotPassword = async function (req: Request) {
```

This doesn't need to be exported as it's only used within this file, it's an internal method.
Cognito is abstracted away by the S4 service so we shouldn't make references to it here, including the error message.

CharlesIC (commented):
A few more suggestions

express/features/support/steps/shared-functions.ts
HarshVasudev: As I have added some step here. Now this file came in the ES linting radar and starting complaining about the usage of any as data type. Replacing any with never has other consequences and on and on :)

HarshVasudev: image

CharlesIC: ok we will sort this out separately, for now we can ignore those warnings

CharlesIC (commented):
This looks fine but commits need to be squashed for this PR

express/src/controllers/sign-in.ts
HarshVasudev: Done.

CharlesIC (approved):

PR #175, opened 2022-11-28
DFA-2281 update stubs for testing

JohnWattsGDS:
Default public for stub response is the default used in to register with Auth.  This means the frontend behaves as though it was live.  If the public key is updated, the stub stores the updated value.

Service name is saved in the stub so proper behaviour is replicated.

These are 'good enough' solutions as the stubs are going to be refactored or replaced at some point.


JohnWattsGDS:
Linting errors are to do with unused stuff in the stubs which we can't really do much about.

CharlesIC (approved):

PR #178, opened 2022-12-06
DFA-2341: Lambda fixes

CharlesIC:
**Correctly set state machine ARNs in lambdas** 

`Ref` returns the ARN for the `AWS::Serverless::StateMachine` type (`Arn` is not available through `GetAtt` according to the docs).

AWS docs: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-statemachine.html#sam-resource-statemachine-return-values

**Set service name for client in the database** 

Use the `service_name` field to store and retrieve the service name.

Set client name to 'integration' since we don't allow to set custom names and we only allow non-prod clients for now.

---

_Linting errors - we haven't appplied Prettier/ESLint to the lambas yet so ignore for now. Reformatting will be done separatelty en masse_


CharlesIC:
Linting errors - we haven't appplied prettier/eslint to the lambas yet so ignore for. Reformatting will be done separatelty en masse

JohnWattsGDS (approved):

PR #181, opened 2022-12-08
DFA-2341: Fixes for the account/service details area of the admin tool

CharlesIC:
**Implement the manage account/client details routes and render the pages with correct parameters**

- Make sure all links work and pass the necessary values to render all the views

- Bring the join private beta routes to the manage account controller from test routes

- Add a return type for the lambdas calling the database API. For now, we're using the AWS SKD types but in the future we should switch to using our own types shared by the API and the frontend (the 'contract)'

  Temporarily Move the `@aws-sdk/client-dynamodb` package to the main `package.json` file as it's now used in multiple projects.

- Don't specify form action path for the forms - they can post to the URLs they're rendered at and we don't need to pass the params as hidden inputs as they are available on the receiving end from URL params on POST

- Check the user is signed in for all manage account routes

- Add new form input types: hidden input, text area, checkboxes, radios and file upload

 **Correctly highlight header and sidebar navigation items** 

Use different names to distinguish the two so on the service details page we can highlight an item in the header as well as in the sidebar.

### DIFF
--- a/express/src/controllers/manage-account.ts
+++ b/express/src/controllers/manage-account.ts
@@ -12,6 +12,9 @@ import AuthenticationResultParser from "../lib/AuthenticationResultParser";
 import {convertToCountryPrefixFormat} from "../lib/mobileNumberUtils";
 import SelfServiceServicesService from "../services/self-service-services-service";
 
+const defaultPublicKey =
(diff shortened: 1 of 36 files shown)


express/src/controllers/manage-account.ts
```diff
@@ -32,9 +35,119 @@ export const listServices = async function (req: Request, res: Response) {
...
+
+export const processPrivateBetaForm = async function (req: Request, res: Response) {
+    const yourName = req.body.yourName;
+    const department = req.body.department;
+    const serviceName = req.body.serviceName;
+    const emailAddress = req.session.emailAddress;
+    const serviceId = req.params.serviceId;
+    const selfServiceClientId = req.params.selfServiceClientId;
+    const clientId = req.params.clientId;
+    const errorMessages = new Map<string, string>();
```
JohnWattsGDS, line 104: I thought we'd banished these Maps and were onto the macro stuff?

CharlesIC: Yes a bit further down `errorMessages: Object.fromEntries(errorMessages)`
This is just to simplify the logic in that function for now. The controllers and validation will need more refactoring but that's for another story. This is fine for the MVP

express/src/controllers/manage-account.ts
```diff
@@ -84,24 +196,22 @@ export const processUpdateServiceForm = async function (req: Request, res: Respo
...
-        res.render("there-is-a-problem.njk");
-        return;
-    }
-
+    const selfServiceClientId = req.params.selfServiceClientId;
+    const clientId = req.params.clientId;
     const s4: SelfServiceServicesService = req.app.get("backing-service");
+
+    // TODO service_name is the db layer leaking into the domain
+    // TODO is this correct? Don't we also need to update the pk:service sk:service entry? We need an updateService call
```
JohnWattsGDS, line 204: The lambda takes care of all of this via a step-function.  The frontend wants to update the client, the backend does that and hides away the complexity that using DynamoDB introduces.  This is probably arse about face because the client name and service name were once the same.  Which probably makes your change that hardcodes 'integration' interesting!

CharlesIC: When updating the service name we need to update the service name field on the client (pk:service sk:client) and on the corresponding pk:service sk:service entry. I don't believe the current handler do that. Changing the service name is out of scope for the MVP so that's fine for now but we'll need to revisit the lambdas and the APIs.

The client name is just for tidyiness as at the moment we don't let the users set or change the client name. They're allowed to have an 'integration' client and in the future a 'production'. We don't have to worry about this until we get to the multi-client functionality.

JohnWattsGDS (approved):

BorisP-GDS (approved):

PR #183, opened 2022-12-09
DFA-2244: Fix visual bugs and dedupe views

CharlesIC:
👉  _Best reviewed by commits_

---

- Use base templates:
  - Use the base form layout in more views
  - Create base templates for:
     - resend security code pages
     - check email pages
     - change phone number pages 

- Simplify back links
   Set a path on the view and generate the back link in the base template
- Don't highlight the "Your account" header navigation item on the service details page
   It should only be highlighted on the "Your account" pages
- Stub service name for sign-in
-  Simplify the security code inputs 
   - Use the same ID for all security code inputs and remove the arg from the form input
   - Dedupe steps in features to submit security codes
   - Rename "otp" to "security code" for consistency
 - Use standard names/IDs for input fields where possible


JohnWattsGDS (approved):
It took a while but 🚀

BorisP-GDS (approved):

PR #184, opened 2022-12-12
DFA-2244: Simplify forms

CharlesIC:
**Decouple forms and POST paths** 

Remove the `formAction` parameter from all views. Forms can post to the path they're rendered on. This allows for less code as any URL params from the GET are also automatically available on the POST and don't need to be passed through as hidden form params.

In addition, the presentation layer should be separate from the routing and page sequencing logic. The views should not be coupled to any specific paths as we want to be able to rearrange them and mount them on different paths as needed. Therefore, hardcoding any paths in views should be avoided.

Further refactoring work will need to remove the back link paths from the views and set them at the point of rendering.

---

- Add JSDoc comments to the CognitoClient to point to the relevant documentation. It is especially helpful to include links to specific API calls used in the client.
https://jsdoc.app/index.html
-  Address unused vars warning in stubs and `app.ts`
-  Fix incorrect back links

### DIFF
--- a/express/src/routes/manage-account.ts
+++ b/express/src/routes/manage-account.ts
@@ -83,6 +83,7 @@ router.post("/change-service-name/:serviceId/:selfServiceClientId/:clientId", ch
 
 router.get("/account/resend-phone-code", checkAuthorisation, (req, res) => {
     res.render("common/resend-security-code.njk", {
+        backLinkPath: "/account/verify-phone-code",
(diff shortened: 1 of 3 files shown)


BorisP-GDS (commented):
Looks good to me

BorisP-GDS (approved):

JohnWattsGDS (approved):
🚀

PR #186, opened 2022-12-13
DFA-2310 Take ownership of sign-in.service.gov.uk

JohnWattsGDS:
Add CloudFormation template to create hosted zone and replicate records currently found in Auth's account.

### DIFF
--- a/backend/domains/production/template.yaml
+++ b/backend/domains/production/template.yaml
@@ -1,0 +1,72 @@
+AWSTemplateFormatVersion: 2010-09-09
+Description: >-
+  DNS configuration for self-service
+
+Transform:
+  - AWS::Serverless-2016-10-31
+
+Resources:
+  SignInServiceHostedZone:
+    Type: AWS::Route53::HostedZone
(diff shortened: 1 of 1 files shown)


backend/domains/template.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+Resources:
+  SignInServiceHostedZone:
+    Type: AWS::Route53::HostedZone
+    Properties:
+      HostedZoneConfig:
+        Comment: Hosted zone for sign-in.service.gov.uk
+      Name: sign-in.service.gov.uk.
+
+  FromAuth:
```
CharlesIC, line 16: A better name for this? `SingInService(Sub)Domains` or something like that?

CharlesIC (commented):
Looks good, just a small comment about naming

backend/domains/template.yaml
JohnWattsGDS: 👍🏻

CharlesIC (approved):

PR #199, opened 2022-12-29
SSE-2385 public key data

JohnWattsGDS:
Use correct client ID
Use public key prepared for Auth Client Registry

### DIFF
--- a/express/src/services/lambda-facade/StubLambdaFacade.ts
+++ b/express/src/services/lambda-facade/StubLambdaFacade.ts
@@ -22,8 +22,9 @@ class StubLambdaFacade implements LambdaFacadeInterface {
         phone: {S: "07700 987 654"}
     };
 
-    publicKey =
+    private DEFAULT_PUBLIC_KEY =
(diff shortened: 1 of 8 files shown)


BorisP-GDS (approved):
Looks good to me

express/src/services/lambda-facade/StubLambdaFacade.ts
```diff
@@ -136,6 +138,10 @@ class StubLambdaFacade implements LambdaFacadeInterface {
     ): Promise<AxiosResponse> {
         return Promise.resolve({} as AxiosResponse);
     }
+
+    resetPublicKey(): void {
+        this.publicKey = this.DEFAULT_PUBLIC_KEY;
+    }
```
CharlesIC, line 144: I don't see why this is needed for the acceptance tests?
If this is the default key, it will be set and returned, won't it? Why would it be different?

The value is not checked in the tests and at the moment we don't even check that the new value is different than the old value when changing the key so we will always see the "You have changed your public key" message" (we only have a test route for updating the key).

Why would the tests only work if the default value is returned?

JohnWattsGDS: If we update the public key, it's stored in the stub.  But because the stub is the most MVP of stubs it only has one user.  When you do the next test, because we've changed it, we have two anchors with the word public key in them so the test fails (you get "Show the full public key").  The path of least resistance was to just reset the key instead of fighting with flaky xpath expressions.

JohnWattsGDS: I'll implement the "... url that starts with ..." step we discussed on Slack.

CharlesIC (approved):

PR #205, opened 2023-01-09
SSE-2310 Self Service development domains

JohnWattsGDS:
Create hosted zones for `development.sign-in.service.gov.uk` and `preview.sign-in.service.gov.uk`.

Create hosted zones to be applied in the development account for the development and preview sub domains.

Update the production hosted zone with NS records pointing to the nameservers in the new hosted zones so that queries will resolve properly.

The development and preview templates have been applied and the nameservers can be checked in Route53 to make sure that the production hosted zone is correct in this PR before it is also applied.

### DIFF
--- a/backend/domains/development/template.yaml
+++ b/backend/domains/development/template.yaml
@@ -1,0 +1,35 @@
+AWSTemplateFormatVersion: 2010-09-09
+Description: >-
+  DNS configuration for self-service development and preview sub domains deployed in development account
+Transform:
+  - AWS::Serverless-2016-10-31
+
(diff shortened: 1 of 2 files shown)


HarshVasudev (dismissed):

CharlesIC:
Development and preview can be one template and stack since they both live in dev

CharlesIC (commented):
Development and preview can be one template and stack since they both live in dev

CharlesIC (approved):
Everything looks fine in AWS 🚀

PR #206, opened 2023-01-16
SSE-2040-update-password-details-on-your-account

HarshVasudev:
- Updating the account details with the password change time.
- A post forgot password confirmation lambda trigger that sends an email using SES
- Domain verification that is needed for SES email sending.
- In sandbox environment receipient also need to be added into the verified email identities.

### DIFF
--- a/backend/cognito/template.yaml
+++ b/backend/cognito/template.yaml
@@ -4,7 +4,94 @@ Description: >-
 Transform:
   - AWS::Serverless-2016-10-31
 
+Parameters:
+  DomainOrEmailIdentityToVerify:
+    Type: String
+    Description: Domain or email that will be created as verified identity.
+  SignInHostedZone:
+    Type: String
+    Description: Deployment specific SignInHostedZone
+
 Resources:
+  DomainIdentityForSES:
+    Type: AWS::SES::EmailIdentity
+    Properties:
+      EmailIdentity: !Ref DomainOrEmailIdentityToVerify
+
+  DevelopmentEmailIdentityRoute53Records:
+    DependsOn:
+      - DomainIdentityForSES
+    Type: AWS::Route53::RecordSetGroup
+    Properties:
+      Comment: "Records created to verify domain ownership with SES"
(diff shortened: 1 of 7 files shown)


backend/cognito/template.yaml
```diff
@@ -4,7 +4,85 @@ Description: >-
...
+    Type: AWS::Serverless::Function
+    Properties:
+      Policies:
+        - Version: '2012-10-17'
+          Statement:
+            - Effect: Allow
+              Action:
+                - 'ses:SendEmail'
+                - 'ses:SendRawEmail'
+              Resource: '*'
```
JohnWattsGDS, line 62: Can we generate the ARN here for the EmailIdentity we've just created instead of a wildcard?

The example format of the ARN is: `arn:aws:ses:us-east-1:888888888888:identity/example.com` so it would need something like:

`!Sub arn:aws:ses:${AWS::Region}:${AWS::AccountId}:identity/${DomainOrEmailIdentityToVerify}` to work.

Where do we get the from address from?  I can't see where we actually pass it in.  And with that in mind, can we look at [restricting the from address](https://docs.aws.amazon.com/ses/latest/dg/sending-authorization-policy-examples.html#sending-authorization-policy-example-from) too while we're at it?

JohnWattsGDS (changes requested):
I've made a request around the policy we create - I've only commented once but it should be applied to all places where the ses:sendEmail permission is added but otherwise looks okay to me.

backend/cognito/template.yaml
HarshVasudev: It is all taken care @JohnWattsGDS. ses:SendEmail was added in two places in the template. Second one was unnecessary. I have removed it. So there is only one place where we need this !Sub arn:aws:ses:${AWS::Region}:${AWS::AccountId}:identity/${DomainOrEmailIdentityToVerify instead of '*'. Which I have tested and working okay after submitting the prod account access for SES.

JohnWattsGDS (approved):
👍🏻

PR #211, opened 2023-01-24
SSE-2386 Public Key not displayed in text area

JohnWattsGDS:
Public Key data should be displayed in the text area used for entering public keys and you never know when you'll need to manually edit the public key you've already successfully stored.

Use correctly sanitised version of public key so comparisons work and the default isn't displayed.

Pass value to template.

Add test.

### DIFF
--- a/express/features/support/steps/client-steps.ts
+++ b/express/features/support/steps/client-steps.ts
@@ -1,5 +1,7 @@
-import {Given} from "@cucumber/cucumber";
+import {Given, Then} from "@cucumber/cucumber";
 import {clickSubmitButton, enterTextIntoTextInput} from "./shared-functions";
+import {strict as assert} from "assert";
+import {TestContext} from "../test-setup";
 
 const VALID_PUBLIC_KEY =
     "-----BEGIN PUBLIC KEY-----\n" +
@@ -29,4 +31,10 @@ Given("the user submits a valid public key with headers", async function () {
 Given("the user submits a valid public key without headers", async function () {
     await enterTextIntoTextInput(this.page, VALID_PUBLIC_KEY_WITHOUT_HEADERS, "serviceUserPublicKey");
     await clickSubmitButton(this.page);
+});
+
+Then("they should see the public key they just entered in the textarea", async function (this: TestContext) {
+    const textArea = await this.page.$("#serviceUserPublicKey");
+    const textAreaContent = await this.page.evaluate(element => element.textContent, textArea);
+    assert.equal(textAreaContent, VALID_PUBLIC_KEY_WITHOUT_HEADERS.replace(/\n/g, ""));
 });
--- a/express/src/controllers/manage-account.ts
+++ b/express/src/controllers/manage-account.ts
@@ -63,19 +63,19 @@ export const showClient = async function (req: Request, res: Response) {
         postLogoutRedirectUrls: client.postLogoutUris.join(" "),
         urls: {
             // TODO changeClientName is currently not used
-            changeClientName: `/change-client-name/${serviceId}/${selfServiceClientId}/${authClientId}?clientName=${encodeURI(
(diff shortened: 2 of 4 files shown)


express/src/routes/testing-routes.ts
```diff
@@ -163,7 +163,8 @@ router.get("/change-public-key/:serviceId/:selfServiceClientId/:clientId", (req,
     res.render("service-details/change-public-key.njk", {
         serviceId: req.params.serviceId,
         selfServiceClientId: req.params.selfServiceClientId,
-        clientId: req.params.clientId
+        clientId: req.params.clientId,
+        values: {serviceUserPublicKey: req.query.publicKey}
```
CharlesIC, line 167: Linter doesn't complain about this hmmm.... 🧐 🤷‍♂️

express/src/controllers/manage-account.ts
```diff
@@ -72,9 +72,7 @@ export const showClient = async function (req: Request, res: Response) {
             changeUserAttributes: `/change-user-attributes/${serviceId}/${selfServiceClientId}/${authClientId}?userAttributes=${encodeURI(
                 client.scopes.join(" ")
             )}`,
-            changePublicKey: `/change-public-key/${serviceId}/${selfServiceClientId}/${authClientId}?publicKey=${encodeURI(
-                client.publicKey
-            )}`,
+            changePublicKey: `/change-public-key/${serviceId}/${selfServiceClientId}/${authClientId}?publicKey=${encodeURI(userPublicKey)}`,
```
CharlesIC, line 75: We should be doing more encoding/encrypting (on all the params) to hide plain data but that's a story for another time....

express/features/support/steps/client-steps.ts
```diff
@@ -30,3 +31,9 @@ Given("the user submits a valid public key without headers", async function () {
     await enterTextIntoTextInput(this.page, VALID_PUBLIC_KEY_WITHOUT_HEADERS, "serviceUserPublicKey");
     await clickSubmitButton(this.page);
 });
+
+Then("they should see the public key they just entered in the textarea", async function () {
+    const textArea = await this.page.$("#serviceUserPublicKey");
+    const textAreaContent = await this.page.evaluate((element: {textContent: string}) => element.textContent, textArea);
```
CharlesIC, line 37:
```suggestion
Then("they should see the public key they just entered in the textarea", async function (this: TestContext) {
    const textArea = await this.page.$("#serviceUserPublicKey");
    const textAreaContent = await this.page.evaluate(element => element.textContent, textArea);
```

Tell the interpreter the type of `this` and then it knows the type of `element` and we don't need the explicit type declaration (`element` is of type `any` but it's not implicit since it's specified by the `Page` type).

express/features/support/steps/client-steps.ts
```diff
@@ -30,3 +31,9 @@ Given("the user submits a valid public key without headers", async function () {
...
+    assert.equal(textAreaContent, VALID_PUBLIC_KEY_WITHOUT_HEADERS.replace(/\n/g, "").replace(/\+/g, " "));
```
CharlesIC, line 38: Why do the `+`s need to be replaced with spaces?
They are valid Base64 characters ([spec](https://www.rfc-editor.org/rfc/rfc4648.txt)). Does the library we use complain about them?

JohnWattsGDS: This was a late in the day brain fart from me - URL encoding must make Express think the +s are actually spaces when you retrieve the query parameter and in my head the Base64 + was interchangeable with a space 🤦🏻‍♂️ 

Anyyywayyyy, it puts the correct text in the text area now so that other than the newlines which still need removing, we don't rejig the sample text.

JohnWattsGDS: TIL `encodeUrlComponent` is what I actually wanted long ago.

CharlesIC (approved):

PR #212, opened 2023-01-24
SSE-2463: Set up infrastructure directory in the repo

CharlesIC:
Set up the infrastructure directory to hold all templates and deployment scripts in one place. This will allow us to reuse the scripts and let SAM builds work with the monorepo structure.

Start by moving the domains and frontend templates.

Add a script to deploy the domain stack for each env. Add account check to output parameters and frontend resources.

**Make hosted zones non-deletable** 

We don't want to accidentally delete the hosted zones when the stacks are updated or accidentally deleted. In fact, we never expect the existing zones to change, so we can mark them with the appropriate metadata and hope SAM won't delete them.

_Changes have been applied to all accounts._


infrastructure/frontend/apply-frontend-infrastructure.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
JohnWattsGDS, line 1: I'm assuming this is solely for development because higher environment deployments will be done using the pipeline stacks? Am I right?

JohnWattsGDS (approved):
🥳

infrastructure/frontend/apply-frontend-infrastructure.sh
CharlesIC: Correct, this is for manual deployment at least for now (though we might be able to resuse the scripts in the workflows and ajust them as needed)

JohnWattsGDS (approved):

HarshVasudev (approved):
Looks good

abdul-gds:
Looks good.

PR #222, opened 2023-01-30
SSE-2418: Manually deploy lambdas to dev

CharlesIC:
**Build and manually deploy the lambdas to dev**

Fix various issues with the lambdas template and address CloudFormation linting warnings.

- Adjust the template to enable SAM/esbuild builds
  Make relative resource paths start from the root of the repo.
  Since we need the npm config from the root of the repo for SAM/esbuild to work, the template needs to tbe built with the base dir set to the root - add a script and npm commands to do that.

- Correctly combine the `Fn::Sub` and `Fn::ImportValue` functions
   The template text was copied to the policy document without the intended substitution.
 _See docs: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-importvalue.html#w2ab1c33c28c41b9_

- Add default values for parameters
  To make deployments easier for now. We might want to change them in the future.

- Use the default values of Metadata and override where necessary
- Use Node.js version 18.x
- Define policy documents as YAML rather than JSON
- Refer to step functions by the ARN (use `!Ref`)

- Move resource definitions around to group related resources together
  Add comments to create sections within the template. If the template gets too big, we can break it up into separate stacks by the groupings.

**Organise lambdas into categories**

Group the lambdas by categories and rename `dynamodb-api` to just `api` since it encapsulates a wider functionality.

Ideally, we'd have a stack per category to keep the templates and stacks small, and to make the deployments more modular, however, that would mean having to set up multiple secure pipelines so we need to balance the number of stacks vs the number of pipelines we want to maintain. We might want to break up the lambdas into multiple stacks/deployments in the future.

---

- Add support for guided deployments to the SAM script


JohnWattsGDS (approved):
It feels a little convoluted using `npm` like that but, meh, it's consistent I suppose and means we don't have to sail around the directory structure to get things done so 👍🏻 let's do this.

PR #223, opened 2023-01-31
SSE-2386 Show public key when there have been input errors

JohnWattsGDS:
Make sure public key can be rendered if there are inline errors

Make text for "Current public key" strong.


BorisP-GDS (approved):

CharlesIC (approved):
Nice 👍

PR #225, opened 2023-02-02
SSE-2463: Show progress when running acceptance tests in GitHub Actions

CharlesIC:
As we've added more acceptance tests, they now take more than 5 minutes to run in GHA. However, cucumber only prints output after all the tests have run, so we don't get any indication of progress. This can't really be fixed due to how cucumber works, but it can be improved by running the tests one feature at a time and printing the output after each.

Additionally, report failed features to the step summary.

For the unit tests, use the dedicated github-actions formatter for jest to get automatic reporting with annotations. This is configured by supplying the dedicated config file.

Adjust the global ts config to include all files not covered by any of the subprojects.

### DIFF
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -18,7 +18,7 @@ jobs:
     timeout-minutes: 5
     steps:
       - name: Run checkov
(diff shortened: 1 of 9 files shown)


HarshVasudev (approved):
Looks good

PR #227, opened 2023-02-03
SSE-2491 Fix Public Key Bug

JohnWattsGDS:
Because we just manipulated the input to return and assumed that the crypto lib would refuse to convert input with junk after the key there was a bug.

We would append anythng on a new line after `-----END PUBLIC KEY-----` to the key we rendered and stored.

This changes that so that we export the converted key to PEM format and then strip that so it works for Auth.

### DIFF
--- a/express/src/lib/publicKeyUtils.ts
+++ b/express/src/lib/publicKeyUtils.ts
@@ -6,17 +6,29 @@ export default function getAuthApiCompliantPublicKey(publicKey: string): string 
     publicKey = publicKey.trim();
     // assume PEM with headers
     try {
-        createPublicKey({key: publicKey, format: "pem"});
-        return publicKey.replace(BEGIN, "").replace(END, "").replace(/\n/g, "").replace(/\r/g, "");
+        return makeAuthCompliant(publicKey);
     } catch (ignored) {}
-    // assume they didn't supply headers
-    const publicKeyWithHeaders = `${BEGIN}\n${publicKey}\n${END}\n`;
 
     try {
-        createPublicKey({key: publicKeyWithHeaders, format: "pem"});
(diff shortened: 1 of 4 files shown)


express/src/lib/publicKeyUtils.ts
```diff
@@ -6,15 +6,33 @@ export default function getAuthApiCompliantPublicKey(publicKey: string): string
...
+        return (
+            createdKey.export({
+                format: "pem",
+                type: "spki"
+            }) as string
+        )
+            .replace(BEGIN, "")
+            .replace(END, "")
+            .replace(/\n/g, "")
+            .replace(/\r/g, "");
```
CharlesIC, line 19: Wrap this in a method

CharlesIC, line 19:
```suggestion
            .replace(/[\n\r]/g, "")
```
FWIW

express/src/lib/publicKeyUtils.ts
```diff
@@ -6,17 +6,29 @@ export default function getAuthApiCompliantPublicKey(publicKey: string): string
...
     }
 }
+
+function makeAuthCompliant(publicKey: string): string {
+    const key = createPublicKey({key: publicKey, format: "pem"});
+    return (
+        key.export({
+            format: "pem",
+            type: "spki"
+        }) as string
```
CharlesIC, line 28: Better to use the `toString()` method on the exported object and let node handle the type conversion if needed, and also make typescript happy knowing the thing will always be a stirng wihtout turning off the type checking.

The methods can be chained as well
```
return createPublicKey({key: publicKey, format: "pem"})
        .export({format: "pem", type: "spki"})
        .toString()
        .replace(BEGIN, "")
        ........
```

CharlesIC (approved):

PR #229, opened 2023-02-07
SSE-2463: Add a script to deploy a SAM stack

CharlesIC:
**Add a script to build and deploy a SAM stack**

Handle common options and scenarios.
Use the script to deploy the different components in the infrastructure directory.
Apply tags to manual deployments.

---

- Organise npm packages
  Remove unused packages.

- Use the latest actions to deploy frontend to PaaS
  Supply additional arguments to make the deployment work with the monorepo and shifted npm dependencies.

- Rename packages to 'self-service'
- Address some linting issues in the templates


infrastructure/deploy-sam-stack.sh
```diff
@@ -0,0 +1,116 @@
...
+  [[ ${#TAGS[@]} -gt 0 ]] && echo "  with tags '${TAGS[*]}'"
+  [[ ${#PARAMS[@]} -gt 0 ]] && echo "  with params '${PARAMS[*]}'"
+fi
+
+export AWS_DEFAULT_REGION=eu-west-2
+
+if [[ $TEMPLATE ]] && $BUILD && $CREDS; then
+  echo "Validating template..."
+  sam validate ${TEMPLATE:+--template "$TEMPLATE"}
+  sam validate ${TEMPLATE:+--template "$TEMPLATE"} --lint
```
JohnWattsGDS, line 91: How does this exit if the template is invalid?

CharlesIC: How do you mean?
If the template file is specified but it doesn't exist, the script exits on line 63.
Actually it will always exist and be valid in this `if` block so the checks on lines 90-91 are not needed.

JohnWattsGDS: `sam validate` is a good command to run, it's just that if the template is deemed to be invalid the script continues anyway.  I'd have thought a check on `$?` at the end of the if block would be useful so that it doesn't attempt to build and invalid template.

JohnWattsGDS: I take it all back ... you've put set -e at the top ...

JohnWattsGDS (approved):
🚀

PR #235, opened 2023-02-15
SSE-2499 - Create 'What do you want to change' page front-end

BorisP-GDS:
SSE-2499 - Create 'What do you want to change' page front-end


CarlyG55:
closing as we are not completing feature development on SSE

PR #238, opened 2023-02-22
SSE-2467 template changes for devplatform

JohnWattsGDS:
Deploying stacks with devplatform requires two extra parameters:

*	`PermissionsBoundary`
*	`CodeSigningConfigArn`

These parameter values are supplied by the secure pipeline deployers and are not applied (`AWS::NoValue`) if the parameter is not passed in.

They need to be added to all Lambdas as they create impicit roles and and custom roles that are created like our `StepFunctionExecutionRole`.  I've used Global for the lambdas but global cannot be used for roles.

### DIFF
--- a/infrastructure/api/sse-api.yml
+++ b/infrastructure/api/sse-api.yml
@@ -15,6 +15,39 @@ Parameters:
     Type: String
     Default: not-a-real-email-address@fake.com
     Description: Email address for private beta request notifications.
+  PermissionsBoundary:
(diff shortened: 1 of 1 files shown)


CharlesIC (approved):

CharlesIC:
Twese two params are required by the Code Pipeline template
https://github.com/alphagov/di-devplatform-deploy/blob/aafe47a8569862e688cc228e650e3ec4d558fb95/sam-deploy-pipeline/template.yaml#L3643-L3644

PR #246, opened 2023-03-14
SSE-2551 devplatform in Staging

JohnWattsGDS:
Write script to capture parameters used to create deployment pipelines.

Write scripts to invoke the devplatform provisioner.sh script and use the parameters we have in the parameter store.  Scripts written for build and development and staging.


infrastructure/ci/dev-platform/shared/save-parameters-to-parameter-store.sh
```diff
@@ -0,0 +1,46 @@
...
+  exit 1
+fi
+
+if [[ "$#" -ne 1 ]]; then
+  echo "$USAGE"
+  exit 1
+fi
+
+BASE_DIR="$(dirname "${BASH_SOURCE[0]}")"
+pushd "$BASE_DIR" > /dev/null
```
CharlesIC, line 18: Do we need to `pushd` in this script? We're not accessing any relative files below.

If we do, I believe `popd > /dev/null` is missing at the end of the script

JohnWattsGDS: 👍🏻 we can do without it.

PR #256, opened 2023-03-21
SSE-2520: Split the delete deployment workflow into two

CharlesIC:
Keep the cleanup functionality in a separate workflow so it can be set to send notifications separately


JohnWattsGDS (approved):
✂️

PR #259, opened 2023-03-21
SSE-2521: Fix issues with building the frontend docker image

CharlesIC:
Fix various issues affecting the docker image build process.

- Use inline YAML instead of JSON
  Use the `AWS::LanguageExtensions` transform to define the ECR repo lifecycle policy in YAML.

- Remove the account check from the frontend template
  We'll need to use this template in different accounts

- Use the latest DI GitHub actions
- Consolidate the domain templates into one template

---

**Include packages and distribution files in the frontend docker image**

Change the workflow to build the docker image after the app has been built.
Pull the distribution files from the frontend app artifact.

Build the image from the root of the repo and include the common dependencies.

---

**Import ECS roles from the frontend stack into the GHA role**

Referring to the roles by the stack prefix is not reliable and has recently stopped working. Use a more robust and secure method by using SSM parameters to reference ECS task roles.

Automatically write the ECS task role ARNs to SSM parameters to be used at deploy time.

---

**Apply AWS best practices for tagging resources**

Use the AWS recommended tagging strategy and naming convention.
Apply tags at the stack level so they would propagate to all managed resources.

AWS docs on tagging:
  - https://docs.aws.amazon.com/IAM/latest/UserGuide/id_tags.html#id_tags_naming
  - https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html#tag-best-practices

---

_**Needs alphagov/di-github-actions#44**_


abdul-gds (approved):
looks good.

PR #262, opened 2023-03-24
SSE-2565 Change email sending to DEVELOPER

JohnWattsGDS:
The verification email should have the EmailSendingAccount set to DEVELOPER with the required properties instead of COGNITO_DEFAULT which doesn't use our email address and has a daily limit on the number of emails that can be sent.

### DIFF
--- a/backend/cognito/template.yaml
+++ b/backend/cognito/template.yaml
@@ -11,7 +11,9 @@ Parameters:
   SignInHostedZone:
     Type: String
     Description: Deployment specific SignInHostedZone
-
+  ExternalId:
+    Type: AWS::SSM::Parameter::Value<String>
+    Default: /self-service/config/cognito/external-id
 Resources:
   DomainIdentityForSES:
     Type: AWS::SES::EmailIdentity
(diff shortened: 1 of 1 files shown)


backend/cognito/template.yaml
```diff
@@ -11,16 +11,17 @@ Parameters:
   SignInHostedZone:
     Type: String
     Description: Deployment specific SignInHostedZone
-
+  ExternalId:
+    Type: AWS::SSM::Parameter::Value<String>
+    Description: Used by Cognito and insisted on by Checkov to avoid the confused deputy problem.
```
CharlesIC, line 16:
```suggestion
```
Chop this line

backend/cognito/template.yaml
```diff
@@ -440,9 +440,12 @@ Resources:
       SmsAuthenticationMessage: '{####} is your GOV.UK One Login security code.'
       MfaConfiguration: OPTIONAL
       EmailConfiguration:
-        EmailSendingAccount: COGNITO_DEFAULT
+        EmailSendingAccount: DEVELOPER
+        SourceArn: !Sub arn:aws:ses:${AWS::Region}:${AWS::AccountId}:identity/${DomainOrEmailIdentityToVerify}
+        From: !Sub 'sign-in@${DomainOrEmailIdentityToVerify}'
```
CharlesIC, line 445:
```suggestion
        From: !Sub 'admin-tool@${DomainOrEmailIdentityToVerify}'
```
This makes more sense IMO

CharlesIC (approved):

PR #263, opened 2023-03-30
SSE-2165: Organise routes by app sections

CharlesIC:
**Organise routes and controllers by app sections**

Ensure each router and controller only touches one section of the app - each should have a single responsibility.

- Add a router for service and base routes and move those routes out of the other routers
- Move handlers around to ensure each router only imports handlers from one controller

_TODO: Move the client routes and handlers into their own files_

**Mount routers on specific paths**

This helps to modularise the app and use composition to combine smaller apps into the self-service system.
We also get more control over the paths and will be able to use this mechanism to mount stubs and testing paths as needed in non-prod environments.

- Remove the mount path from routes. Routers specify paths relative to the path they've been mounted on
- Make redirects in routers relative (without the leading '\')
- Mount the testing routes on a path as well to avoid cross-contamination with the main app (some pages will need to be moved later)

**Use`Router.use()` to enforce the authoriser on routes**

The authoriser can be used once before all the routes that need to use it, instead of repeating it for each route.

**Use `Router.route()` to avoid repeating routes**

Use the `Router.route()` method to set multiple handlers on the same route to avoid duplication and reduce the scope for errors.

_Express docs:
https://expressjs.com/en/4x/api.html#router.route_

**Use the `RequestHandler` type from Express for middleware functions**

Avoid the linting error caused by the unused type variables, and don't replicate the type from the framework.

---

- Update packages
- Update the owners file


BorisP-GDS (approved):
Looks good to me

PR #266, opened 2023-03-31
SSE-2165: Separate Clients from Services

CharlesIC:
**Create a router and controller for clients**

- Split the services controller into two to ensure they each have a single responsibility
- Mount the clients routes from the services routes as they are dependent on the parent service
- Pull the `serviceId` from the request context and check it's not null where needed

**Add a utility function to get optional values**

The function takes a value or an object and a set of property names, and returns the values if they are present or throws an exception if values are undefined.

This simplifies checking for nulls at runtime and can be used where an optional value is not expected.

The function can be extended in the future to include a custom handler or a custom error message.

Register the function on `globalThis` and import the module in `app.ts` so it is available globally and doesn't need to be explicitly imported.

**Add context to requests**

- Extend the express request interface to include an object with optional pre-defined properties.
- Extend the express application to initialise the context object with an empty value.

This works the same way as the `req.session` object but is intended to store information relevant to a single request-response cycle rather than the entire user's session.

This is useful when we want to mount an app on a sub-path that includes a parameter, since `req.params` won't be available to the child app. The parent app must set the params it wants to pass to the child app.

Express docs recommend setting properties directly on the request object but since we use TypeScript we need to pre-define the object and the properties it will hold to avoid errors and get typechecking.

_Express reference:
https://expressjs.com/en/4x/api.html#router.param_

---

**Tech debt**

- Move the session data declaration to its own file to allow reuse
  Import the declaration as a module in the tests that need it instead of repeating the declaration.

- Add middleware to host error handling functions
- Tidy up session storage and move it to config
- Import `Router` from express directly
- Move the root handler to the base router
- Shift things around in `app.ts` for better organisation


BorisP-GDS (approved):
Looks good to me

PR #269, opened 2023-03-31
SSE-2576 API Gateway logging stacks

JohnWattsGDS:
Add scripts used to create the api gateway logging stacks to the repo.

The script has not been run on development as a differently named stack already exists.


abdul-gds (approved):
looks good

PR #270, opened 2023-04-04
SSE-2559 - Move routes used in the main app for One Login Admin out of testing routes

BorisP-GDS:
Move routes used in the main app for One Login Admin out of the testing routes

### DIFF
--- a/express/src/controllers/clients.ts
+++ b/express/src/controllers/clients.ts
@@ -9,7 +9,7 @@ export const showClient: RequestHandler = async (req, res) => {
     // TODO make S4 instances static in all controllers
     const serviceId = req.context.serviceId;
     const s4: SelfServiceServicesService = req.app.get("backing-service");
-    const client = (await s4.listClients(nonNull(serviceId), req.session.authenticationResult?.AccessToken as string))[0];
+    const client = (await s4.listClients(nonNull(serviceId), nonNull(req.session.authenticationResult?.AccessToken)))[0];
     const selfServiceClientId = client.dynamoServiceId;
     const authClientId = client.authClientId;
     const serviceName = client.serviceName;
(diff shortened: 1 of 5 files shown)


express/src/services/lambda-facade/StubLambdaFacade.ts
```diff
@@ -140,6 +153,10 @@ class StubLambdaFacade implements LambdaFacadeInterface {
     ): Promise<AxiosResponse> {
         return Promise.resolve({} as AxiosResponse);
     }
+
+    marshallArray(uris: string[]): Record<string, unknown> {
+        return {L: marshall(uris)};
+    }
```
CharlesIC, line 159: Import and use this method instead of the custom function - the SDK method automatically converts JavaScript objects to the list type `AttributeValue`
```
import {convertToAttr} from "@aws-sdk/util-dynamodb";
```

express/src/services/lambda-facade/StubLambdaFacade.ts
```diff
@@ -61,14 +65,23 @@ class StubLambdaFacade implements LambdaFacadeInterface {
         serviceId: string,
         selfServiceClientId: string,
         clientId: string,
-        updates: Record<string, string>,
+        updates: Record<string, unknown>,
```
CharlesIC, line 68:
```suggestion
        updates: Record<string, string | string[]>;,
```
Use the specific type and the `convertToAttr` SDK method to automatically handle type conversions and avoid type warnings.

CharlesIC (commented):
Some changes are needed

express/src/controllers/clients.ts
```diff
@@ -158,3 +158,120 @@ export const processChangeServiceNameForm: RequestHandler = async (req, res) =>
     req.session.serviceName = newServiceName;
     res.redirect(`/services/${serviceId}/clients`);
 };
+
+export const showProcessChangePublicKeyForm: RequestHandler = async (req, res) => {
+    res.render("service-details/change-public-key.njk", {
+        serviceId: req.params.serviceId,
```
CharlesIC, line 164:
```suggestion
        serviceId: req.context.serviceId,
```
Service ID is avaialble on the request context and won't be available on the params since the clients controller is mounted on `/clients/:clientId`

CharlesIC (approved):

PR #278, opened 2023-04-20
SSE-2598 - Refactor sign in mobile code resend

BorisP-GDS:
Refactor sign in mobile code resend


BorisP-GDS:
Closing as the story was replaced by SSE-2609 , and the work for it is done in https://github.com/alphagov/di-onboarding-self-service-experience/pull/305 , so this one is no longer needed

PR #280, opened 2023-04-24
SSE-2489: Add typechecking to tests

CharlesIC:
- Rename acceptance tests steps for consistency

- Set up typechecking for tests

  Add `tsconfig` for tests, so they get typechecking and other checks
  Set up config pattern for jest and add support for relative paths
  Set up the `src` directory as a root and use relative paths in tests


BorisP-GDS (approved):
Looks good to me

PR #282, opened 2023-04-25
BAU: Use default exports

CharlesIC:
- Organise files in express; use consistent naming
- Use default exports to expose functionality where it makes sense
- Move session data to its own file and import
- Add the `NOTICE.md` file for 3rd party licences


BorisP-GDS (approved):
Looks good to me

PR #284, opened 2023-04-25
BAU: Update validators

CharlesIC:
- Add a validator for allowed domains to encapsulate the logic for verifying emails
- Use default exports in validators
- Streamline the logic and trim down the code
---
- Common passwords validator - export a function instead of a class
- Public key utils - extract the key content and wrap it in headers to ensure whitespace and newlines inside the key are removed


BorisP-GDS (approved):
Looks good to me

PR #285, opened 2023-04-25
BAU: Update middleware validator functions

CharlesIC:
- Use default exports in middleware validator functions
- Remove unnecessary async
- Streamline the logic and trim down the code
- Rename functions to be consistent with controllers
- Reduce nesting in functions by returnig early


BorisP-GDS (approved):
Looks good to me.

PR #289, opened 2023-04-27
SSE-2622 Split out email identity creation

JohnWattsGDS:
Separate creation of email identity into different stack with script to deploy it and check whether id gets verified.


abdul-gds (approved):
looks fine

PR #291, opened 2023-04-28
BAU: Update Cognito and Lambda clients

CharlesIC:
Tidy up clients for Cognito and Lambda
Add methods to send requests and use them in all functions for consistent usage
Remove unnecessary return types and async, or return void if the output is not used

Add config for env vars to manage them all in one place and set defaults as needed

### DIFF
--- a/express/src/services/lambda-facade/LambdaFacade.ts
+++ b/express/src/services/lambda-facade/LambdaFacade.ts
@@ -3,78 +3,59 @@ import {QueryCommandOutput} from "@aws-sdk/client-dynamodb";
 import axios, {Axios, AxiosResponse} from "axios";
 import {OnboardingTableItem} from "../../../@types/OnboardingTableItem";
 import {Service} from "../../../@types/Service";
+import {lambda} from "../../config/environment";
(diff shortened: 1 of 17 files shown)


BorisP-GDS (approved):
Looks good to me

express/src/services/lambda-facade/LambdaFacade.ts
```diff
@@ -3,78 +3,59 @@ import {QueryCommandOutput} from "@aws-sdk/client-dynamodb";
...
-            headers: {
-                "authorised-by": accessToken
-            }
-        });
+
+        await this.post("/new-service", JSON.stringify(body), accessToken);
     }
 
-    async generateClient(service: Service, authenticationResult: AuthenticationResultType): Promise<AxiosResponse> {
+    generateClient(service: Service, authenticationResult: AuthenticationResultType): Promise<AxiosResponse> {
```
JohnWattsGDS, line 43: I don't normally like storycreep but this would have been a good place to slot in the API key stuff.

CharlesIC: Will do this separately as this was already a big PR

PR #292, opened 2023-05-02
SSE-2534 api deploy workflow

abdul-gds:
# Code Changes

- add workflow for API deployment


.github/workflows/upload-api-secure-pipelines.yml
```diff
@@ -0,0 +1,44 @@
+name: Upload API to secure pipelines
+
+on:
+#   workflow_dispatch:
```
JohnWattsGDS, line 4: are these to be uncommented?

JohnWattsGDS:
We could probably do with squishing all the commits.

.github/workflows/upload-api-secure-pipelines.yml
abdul-gds: yes will uncomment

CharlesIC (approved):

PR #293, opened 2023-05-03
BAU: Tidy up controllers

CharlesIC:
**Add utility middleware to create request handlers**

- Introduce simple utility methods to streamline the creation of middleware functions
- Use the functions to define the type of the handlers for better type checking
- Use the utility functions in the middleware library
- Add a utility function to create a middleware function to render a view
  Use the function in routes

**Tidy up controllers**

- Use nonNull instead of type assertion (`as <type>`) to ensure values are actually present at run-time
- Try-catch exceptions around methods that throw them
- Use `RequestHandler` type for controller middleware functions
- Use arrow functions so the function type can be specified
- Use the render helper function
- Return `res.render` and `res.redirect`
- Remove unnecessary `async`

---
- Use the password validator when changing the password in the account
- Use `convertToAttr` in `dynamo-db` client
  Use `convertToAttr` instead of marshall as it correctly converts to the required Cognito types.
  Remove custom implementation of the same functionality.

### DIFF
--- a/acceptance-tests/features/your-account.feature
+++ b/acceptance-tests/features/your-account.feature
@@ -90,12 +90,13 @@ Feature: A page where users can view and change the details associated with thei
     Scenario: The user tries to change their current password and does not enter any value into any of the two fields
       When they click the Confirm button
       Then the error message "Enter your current password" must be displayed for the current password field
+      Then the error message "Enter your new password" must be displayed for the new password field
 
-    Scenario: The user tries to change their current password and does not enter any value into Enter a new password field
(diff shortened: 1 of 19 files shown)


JohnWattsGDS:
I'm trying to work on https://govukverify.atlassian.net/browse/SSE-2623 and I'm guessing that the changes to the password stuff are going to make my life much more difficult than it should be.  Can you just leave those out until I've fixed the reset password stuff?

CharlesIC:
I don't expect conflicts but I'll wait with merging/rebasing until SSE-2623 is done

BorisP-GDS (approved):
Looks good to me.

PR #297, opened 2023-05-04
BAU add missing permissions to StepFunctionExecutionPolicy

JohnWattsGDS:
More permission are needed for the step function to be able to log to a specific log group.


abdul-gds (approved):
looks good. Thanks

PR #308, opened 2023-05-10
SSE-2620 - Update User Attributes Format

BorisP-GDS:
Update User Attributes Format

### DIFF
--- a/express/src/controllers/clients.ts
+++ b/express/src/controllers/clients.ts
@@ -24,7 +24,7 @@ export const showClient: RequestHandler = async (req, res) => {
         serviceName: serviceName,
         updatedField: req.session.updatedField,
         redirectUrls: redirectUrls,
-        userAttributesRequired: client.scopes.join(", "),
+        userAttributesRequired: client.scopes,
         userPublicKey: userPublicKey,
         postLogoutRedirectUrls: client.postLogoutUris.join(" "),
         urls: {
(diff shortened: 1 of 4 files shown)


express/src/controllers/clients.ts
```diff
@@ -23,7 +36,7 @@ export const showClient: RequestHandler = async (req, res) => {
         serviceName: serviceName,
         updatedField: req.session.updatedField,
         redirectUrls: redirectUrls,
-        userAttributesRequired: client.scopes.join(", "),
+        userAttributesRequired: userAttributes.join("<br> "),
```
CharlesIC, line 39:
```suggestion
        userAttributesRequired: userAttributes.join("<br> "),
```
This should be generated in the view, not in the controller. The controller doesn't know about the presentation layer or HTML.

express/src/controllers/clients.ts
```diff
@@ -15,6 +15,19 @@ export const showClient: RequestHandler = async (req, res) => {
...
+        userAttributes.push("Email address");
+    }
+
+    if (client.scopes.includes("phone")) {
+        userAttributes.push("Phone number");
+    }
+
+    if (client.scopes.includes("offline_access")) {
+        userAttributes.push("Offline access");
+    }
```
CharlesIC, line 30:
```suggestion
    if (client.scopes.includes("email")) {
        userAttributes.push("Email address");
    }

    if (client.scopes.includes("phone")) {
        userAttributes.push("Phone number");
    }

    if (client.scopes.includes("offline_access")) {
        userAttributes.push("Offline access");
    }
```
The controller should return the original array. The view should decide what content to show for each token (e.g. print "Offline access" for the `offline_access` token)

express/src/controllers/clients.ts
```diff
@@ -206,11 +219,16 @@ export const processChangeRedirectUrlsForm: RequestHandler = async (req, res) =>
...
+    let email = false;
+    let phone = false;
+    let offline_access = false;
+
+    if (req.query.userAttributes) {
+        const userAttributes = req.query.userAttributes?.toString().split(" ");
+        email = userAttributes.includes("email");
+        phone = userAttributes.includes("phone");
+        offline_access = userAttributes.includes("offline_access");
+    }
```
CharlesIC, line 231: The user attributes are checkboxes, the controller should return the names of the checked boxes. The govuk checkboxes components handles this.

CharlesIC (approved):

PR #310, opened 2023-05-11
Development Stacks go!

JohnWattsGDS:
Enable optional prefixing of stacks for deployments to development environment


README.md
```diff
@@ -15,13 +15,37 @@ Make sure you have an appropriate version of node. 14 is the current LTS version
 brew install nvm
 # follow the instructions about adding some commands to your shell, creating ~/.zshrc or whatever your shell uses if necessary
 . ~/.zshrc
-nvm install 14.20.0
-nvm use 14.20.0
+nvm install 18
+nvm use 18
```
CharlesIC, line 19: The line above the script says we use 14 but actually it's 18 for Lambdas I think.
Locally it doesn't matter really, we should use the latest and we transpile to whatever version is supported on the server (target is `ES2022`).

Mind you not everyone uses `nvm` (I use `nodenv`).

backend/cognito/template.yml
```diff
@@ -14,8 +14,13 @@ Parameters:
     Type: String
     Default: ""
     Description: The ARN of the code signing configuration resource to use [Secure Pipeline]
+  ExportPrefix:
+    Type: String
+    Default: ""
+    Description: Used to differentiate stacks in development account
```
CharlesIC, line 20:
```suggestion
  ExportNamePrefix:
    Type: String
    Default: self-service
    AllowedPattern: ^.*[^-]$
    Description: Prefix to use when importing or exporting values from other stacks
```
We might want to use this for any purpose.
Also we can get rid of the condition by requiring the prefix to not end with a hyphen and have a sensible default value that can be used for both exporting and importing values from other stacks.

backend/cognito/template.yml
```diff
@@ -14,8 +14,13 @@ Parameters:
...
 
 Conditions:
+  UseExportPrefix: !Not [ !Equals [ !Ref ExportPrefix, "" ] ]
```
CharlesIC, line 23:
```suggestion
```

backend/cognito/template.yml
```diff
@@ -616,10 +621,10 @@ Outputs:
     Description: User Pool generated for this stack
     Value: !Ref UserPool
     Export:
-      Name: CognitoUserPool
+      Name: !If [ UseExportPrefix, !Sub "${ExportPrefix}-CognitoUserPool", "CognitoUserPool" ]
```
CharlesIC, line 624:
```suggestion
      Name: !Sub ${ExportNamePrefix}-CognitoUserPool"
```
And the same for all the exports

README.md
JohnWattsGDS: I went with 18 because of the lambdas.  I'll stick in something about using the latest version and transpiling but 18 seems a reasonable default for the docs for someone to follow. I can mention other node version managers are available but I feel like this the thing one follows when they've pulled the repo and don't know where to start.  As such, basic uncomplicated instructions are best because whoever is using them is probably already drinking from a firehose.

Follow those instructions blindly and you can get a working setup to test things out in a few minutes.  Until you use a new language feature the ink hasn't dried on and 18 is no longer working 😆

backend/cognito/template.yml
JohnWattsGDS: Fancy regex 😄 .

CharlesIC (approved):

PR #324, opened 2023-05-26
SSE-2679 - Update the dashboard to indicate that OpenID is always a required attribute

BorisP-GDS:
Update the dashboard to indicate that OpenID is always a required attribute

### DIFF
--- a/express/src/views/clients/client-details.njk
+++ b/express/src/views/clients/client-details.njk
@@ -6,6 +6,7 @@
 
 {% set userAttributes %}
   {% for item in userAttributesRequired %}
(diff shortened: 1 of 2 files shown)


express/src/views/clients/client-details.njk
```diff
@@ -5,6 +5,7 @@
 {% set pageTitle = "Client details" %}
 
 {% set userAttributes %}
+  OpenID<br>
```
CharlesIC, line 8:
```suggestion
  {% if item == "email" %}Email address<br>{% endif %}
```
Let's do the same for OpenID

CharlesIC (approved):

PR #327, opened 2023-06-01
SSE-2643: Enable AWS WAF for frontend load balancer

abdul-gds:
# Code Changes

- add WAF resources for adding security
- AWS-CRS
- Bad-Inputs
- SQLInject-RuleSet
- RateBased-CountIpRule
-web Acl logging


infrastructure/ci/dev-platform/frontend-template.yml
```diff
@@ -339,6 +339,196 @@ Resources:
         Fn::ImportValue:
           !Sub ${VpcStackName}-VpcId
 
+
+  frontendWebAcl:
```
CharlesIC, line 343:
```suggestion
  FrontendWebAcl:
```
And the same for the other new resources to use the same naming convention as we use in the rest of the template

CharlesIC (commented):
Small issue with naming resources and the conflict needs to be resolved, otherwise the PR looks good

infrastructure/ci/dev-platform/frontend-template.yml
abdul-gds: fixed

abdul-gds:

PR #338, opened 2023-06-13
SSE-2643 - add Self-service  Main WAF Web ACL

abdul-gds:
# Code Changes
 WAF resources for adding extra security
- AWS-CRS
- Bad-Inputs
- SQLInject-RuleSet
- RateBased-CountIpRule
- web Acl logging


abdul-gds:

PR #339, opened 2023-06-14
SSE-2685: Deploy backend components through secure pipeliens

CharlesIC:
**Deploy backend components to secure pipelines**

- Modify the API workflow to account for the filepath change
- Add jobs to build and upload the API to secure pipelines
- Upload the secure pipelines workflow to use reusable workflows for each component

**Manage secrets and deployment parameters**
- Add the Auth API key to the deployment parameters script
- Add methods to handle secrets
- Add an array with secret names that need to be set in the account
- Make the script methods more reusable

**Update API template**

Address CloudFormation linting issues
Use parameters and imports consistently
Move the API template and resources to `backend/api`
Remove custom API deployment script (new one coming in another PR)

**SSE-2636: Remove Auth Lambda custom environment encryption key**

Lambda encrypts environment variables at rest by default, using the default Lambda service key.
Since we don't have a need to customise the key rotation schedule, we don't need to use a custom key for this purpose.


abdul-gds (approved):
looks good

rafal-surowiec-co (approved):

PR #343, opened 2023-06-20
SSE-2685: Secure pipelines - await deployment to complete in GHA

CharlesIC:
**Await CodePipeline execution after uploading SAM packages to secure pipelines**

Use the `secure-pipelines/deploy-application` action that will cause the GitHub Actions job to keep spinning for each component until its pipeline finishes with a success or failure state.

The deployer roles need to be given permission to list their pipelines' executions.

---

- Show pipeline URL in the GitHub Actions UI
- Update `PipelineEnvironentNameEnabled` param as the typo has been fixed
- Don't export region in deployment scripts
  The `gds-cli` exports the region environment variables as of [version 5.75.0](https://github.com/alphagov/gds-cli/releases/tag/v5.75.0)
  There's no longer a need to set this manually.
- Update dev script
  Show all stacks with a partial prefix
- Update packages
  Update session storage and data interface to support improved typechecking in the latest version of connect-dynamodb

### DIFF
--- a/.github/workflows/build-api.yml
+++ b/.github/workflows/build-api.yml
@@ -19,7 +19,8 @@ jobs:
     outputs:
       artifact-name: ${{ steps.build.outputs.artifact-name }}
     steps:
-      - uses: alphagov/di-github-actions/sam/build-application@f1aa97d64bf1be3a6226e0319c597bf4b28c19c8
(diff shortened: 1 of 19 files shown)


abdul-gds (approved):
looks good

PR #346, opened 2023-07-04
Changes for the frontend to use a session secret retrieved from AWS S…

MartinH-DI:
Changes for the frontend to use a session secret retrieved from AWS Secrets  Manager

### DIFF
--- a/express/src/config/environment.ts
+++ b/express/src/config/environment.ts
@@ -3,6 +3,8 @@ export const googleTagId = process.env.GOOGLE_TAG_ID;
 export const showTestBanner = process.env.SHOW_TEST_BANNER === "true";
 export const region = process.env.AWS_REGION || process.env.AWS_DEFAULT_REGION;
 
+export const sessionSecret = process.env.SESSION_SECRET;
+
 export const cognito = {
     client: process.env.COGNITO_CLIENT ?? "CognitoClient",
     userPoolId: process.env.COGNITO_USER_POOL_ID,
--- a/express/src/config/session-storage.ts
+++ b/express/src/config/session-storage.ts
@@ -1,7 +1,7 @@
 import connect from "connect-dynamodb";
 import {RequestHandler} from "express";
 import session, {SessionData} from "express-session";
-import {region, sessionStorage} from "./environment";
+import {region, sessionSecret, sessionStorage} from "./environment";
 import {DynamoDB} from "@aws-sdk/client-dynamodb";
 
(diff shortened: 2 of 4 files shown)


express/src/config/session-storage.ts
```diff
@@ -10,9 +10,12 @@ function getSessionStorage(): RequestHandler {
     if (sessionStorage.tableName) {
         console.log("Using DynamoDB session storage");
         const DynamoDBStore = connect<SessionData>(session);
+        console.log(`********* Retrieving Session Storage. Secret is "${sessionSecret}". *********`);
```
CharlesIC, line 13:
```suggestion
```
We do not want to log a secret....

express/src/config/environment.ts
```diff
@@ -3,6 +3,8 @@ export const googleTagId = process.env.GOOGLE_TAG_ID;
 export const showTestBanner = process.env.SHOW_TEST_BANNER === "true";
 export const region = process.env.AWS_REGION || process.env.AWS_DEFAULT_REGION;
 
+export const sessionSecret = process.env.SESSION_SECRET || "ThisIsNotASecret!OhNoNotAtAll*NotInAnyWay";
```
CharlesIC, line 6:
```suggestion
export const sessionSecret = process.env.SESSION_SECRET;
```
The default should be undefined. In `express/src/config/session-storage.ts`, if dynamo storage is selected but no secret is provided, that should be an error.

express/src/config/session-storage.ts
MartinH-DI: Sure; this was a debugging statement that I should have removed.

express/src/config/environment.ts
MartinH-DI: Sure; I'll remove the default.

express/src/config/session-storage.ts
```diff
@@ -11,8 +11,10 @@ function getSessionStorage(): RequestHandler {
         console.log("Using DynamoDB session storage");
         const DynamoDBStore = connect<SessionData>(session);
         return session({
-            secret: "keyboard cat",
+            secret: sessionSecret as string,
```
CharlesIC, line 14:
```suggestion
            secret: nonNull(sessionSecret),
```
Type assertions must be avoided as they bypass type checking. We can use the helper method to ensure the value is not null.

CharlesIC (approved):

CharlesIC (approved):

PR #351, opened 2023-07-13
SSE-2740: Add CloudFront distribution and WAF for Admin Tool frontend

abdul-gds:
# Code Changes

- add WAF for cloudfront
- add Cloudfront for frontend


abdul-gds:

PR #354, opened 2023-07-17
SSE-2638: Add an option to make the backend API private

CharlesIC:
**Make the backend API private by default**

When the frontend runs in the VPC, it can only communicate with the backend API Gateway via a VPC endpoint, meaning the endpoint must be set to private and the VPC subnets needs to be provided.

When running the frontend locally, we need the API to be public, so add a param to enable that and not set the private endpoint config.

Note that the API URL is different when executed through a VPC endpoint, hence the conditional on the output value.

Use a REST API as it's the only one that supports private endpoints.

**Update the frontend template**

- Set up ECS service
- Add container logs
- Add ECS params & env vars
- Add Cognito permissions
- Set up container healtcheck
- Don't assign public IP to frontend service as we'll be using API Gateway and CloudFront to get public access
- Import subnet list from the VPC stack
- Update task role permissions

- Use HTTP API for frontend API Gateway

- Add API Gateway access log settings

- Configure multiline logging for the awslogs driver

---

**Set the session secret automatically**

If the secret is not present, generate a new UUID value.

### DIFF
--- a/backend/api/api.template.yml
+++ b/backend/api/api.template.yml
@@ -1,8 +1,6 @@
 AWSTemplateFormatVersion: "2010-09-09"
 Description: DI Self-Service API
-Transform:
-  - AWS::Serverless-2016-10-31
-  - AWS::LanguageExtensions
+Transform: [ AWS::LanguageExtensions, AWS::Serverless-2016-10-31 ]
 
 Parameters:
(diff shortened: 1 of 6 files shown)


abdul-gds (approved):
looks good.

PR #360, opened 2023-07-24
SSE-2740: Add a CloudFront distribution to the frontend template

CharlesIC:
- Add CloudFront config with caching and access logging

- Add session secret value to frontend environment
- Ensure access logs buckets are globally unique by appending an ID
- Output frontend URL


abdul-gds (approved):
looks good

PR #364, opened 2023-07-31
SSE-2769: Use custom domain for admin tool

CharlesIC:
**Create a VPC template and stack**

Move the VPC to the networking config template
Hardcode Firewall subnet IDs for now (waiting for a change to the VPC template to export those IDs)

**Add TLS certificates**

Add certs to use when setting up an HTTPS endpoint for the frontend load balancer
Export the admin domain values

**Use a custom domain for frontend**

Make the frontend ALB public and allow it to receive HTTPS traffic
Remove redundant networking components


abdul-gds (approved):
looks good

rafal-surowiec-co (approved):

abdul-gds:
should not we move  '`AccessLogsBucket`'  this in our config template, as  It create problem when we need to delete stack - non-empty bucket are not removed
https://github.com/alphagov/di-onboarding-self-service-experience/blob/4786b1e792e[…]1e0f6625b8d760477/infrastructure/frontend/frontend.template.yml

PR #368, opened 2023-08-04
SSE-2678:Added Checkov Hook to Infrastructure Template.

DStumpGDS:
Deployed to Development only (Karl will implement all Accounts).


CharlesIC (dismissed):

CharlesIC (changes requested):
Changes look fine
Commits need to be squashed

keith-hart:
Have checked and squashed commits with David

CharlesIC (approved):

PR #370, opened 2023-08-09
Add custom Cognito attribute for tracking the signup status.

MartinH-DI:
Add custom attribute to allow signup to be restarted from the place it last reached. Per Cognito rules, it cannot be set as required or with any default value, so correct population must be enforced in code.


keith-hart (approved):

abdul-gds (approved):
looks fine

MartinH-DI:
Closing as now subsumed within later work on the signup pathway.

PR #374, opened 2023-08-14
add KMSApiEnabled vpc endpoint

abdul-gds:
set `KMSApiEnabled` VPC template param


abdul-gds:
code was updated by other PR

PR #375, opened 2023-08-15
SSE 2781 fix create account issues when navigating back during process

DStumpGDS:
For Martin to Preview (before final testing).

### DIFF
--- a/express/src/controllers/register.ts
+++ b/express/src/controllers/register.ts
@@ -1,8 +1,4 @@
-import {
-    CodeMismatchException,
-    NotAuthorizedException,
-    UsernameExistsException
-} from "@aws-sdk/client-cognito-identity-provider";
+import {CodeMismatchException, NotAuthorizedException, UsernameExistsException} from "@aws-sdk/client-cognito-identity-provider";
 import {randomUUID} from "crypto";
 import {RequestHandler} from "express";
 import {Service} from "../../@types/Service";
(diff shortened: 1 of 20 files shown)


DStumpGDS:
Incorrectly released (should still be in Draft).

MartinH-DI:
This solution, as drafted, looks a good fit for the requirement. I suggest test in dev environment against the defect tickets.

CharlesIC:
May I suggest investigating two possible solutions before introducting the overhead and complexity (albeit small) of a custom attribute.

1. The issue we're trying to solve is the errors we get when a user tries to use an account that has not completed registration. Interrogating the exception returned from Cognito should provide enough information to handle it - the Cognito SKD returns specific exceptuion types for different scenarios (the list can be referenced from the [Cognito SDK documentation](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/cognito-identity-provider/) for each command). For instance, the `[AdminInitiateAuthCommand](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/cognito-identity-provider/command/AdminInitiateAuthCommand/)` will throw `[UserNotConfirmedException](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-cognito-identity-provider/Class/UserNotConfirmedException)` if the account has not been verified. This approach requires understanding the SDK calls we make to Cognito and what exceptions they may return.

2. Cognito already has all the registration state information so we can Investigate it by making SDK calls to get the information we need. For instance, the `AdminGetUserCommand` command returns an object that can be inspected to check if the user's email or MFA option has been verified, etc., negating the need for a custom attribute.

The above should be investigated in the first instance as they should cover most of the errors caused by incomplete registration. For the outstanding cases, if there are any, we might consider custom attributes.

If we do go the custom attribute route, I'd recommend a boolean for each state "flag" we want, for instance "email_verified", "phone_verified", etc rather than an enum for each stage of the journey as that makes the journey infexible and difficult to make changes to the flow.

I strongly suggest we should avoid adding custom attributes for the purpose of holding state information as that information can already be extracted from Cognito, and doing so adds a custom layer of complexity that reduces flexibility and needs to be managed manually, and is difficult to test. If we do decide to go that route, however, a set of boolean flags is a safer way to go.

DStumpGDS:
In response to Karol's comment above (Thanks Karol!).

We did investigate the solution you suggested but found it did not supply enough 'Stage' Information to allow us to capture all states (and you allude to that yourself) so this would mean having a combination of checking State via the AWS functionality as well as checking local State Flags as well which I strongly felt was prone to even more complexity and possible error situations then a simple State Status.

In response to your suggestion of Boolean Flags this would complicate the checking of Status greatly since one would have to 'scan' all the Flags to determine the current state whereas a single State Attribute allows a quick comparison with known States.

Finally in response to a single variable causing inflexibility the Custom Attribute is a String which means there is a degree of flexibility in that the State can be 'added to' rather then have a fixed set of stages.

Just to confirm and clarify unless there is any further objections raised the currently proposed solution is the path we are currently following.

keith-hart (dismissed):

express/src/controllers/register.ts
```diff
@@ -7,24 +7,52 @@ import {domainUserToDynamoUser} from "../lib/models/user-utils";
...
+            const signUpStatus: SignupStatus = await s4.getSignUpStatus(emailAddress);
+
+            console.log("In UserNameExistException");
+
+            if (!signUpStatus.hasStage(SignupStatusStage.HasEmail)) {
+                console.log("Processing No HasEMail");
+                return res.redirect("resume-before-password");
+            }
+            if (!signUpStatus.hasStage(SignupStatusStage.HasPassword)) {
+                console.log("Processing No HasPassword");
```
rafal-surowiec-co, line 37: All console logs need to be removed

express/src/controllers/register.ts
```diff
@@ -223,3 +261,27 @@ export const accountExists: RequestHandler = (req, res) => {
...
+    const userPassword = nonNull(req.body["password"]);
+    console.log("User Name => " + userName);
+    console.log("User Password => " + userPassword);
+
+    const response = await s4.submitUsernamePassword(userName, userPassword);
+    req.session.cognitoSession = response.Session;
+    req.session.authenticationResult = response.AuthenticationResult;
+    //req.session.mobileNumber = await s4.getMobilePhoneNumber(userName);
+
+    // Keith Hart has requested even if User has already entered their phone number we 'resume' from re-entering their Phone number again.
```
rafal-surowiec-co, line 279: I think these explanations are better put on the ticket and not in the code. Commented code should be removed.

DStumpGDS:
This Pull Request is now being closed without being progressed because it has been superseded by PR SSE-2781-Squashed and that PR has now been merged into Main Branch.

PR #378, opened 2023-08-17
SSE-2521: Deploy admin tool frontend from GitHub Actions to dev

CharlesIC:
**Update dev deployment workflow**

Update the workflow for the latest changes in frontned components (public ALB and a DNS record with a custom domain)
Set correct parameters and values
Update permissions for the GitHub Actions dev role

**Move ELB access logs bucket to the network config template**

Use one bucket for all ALB logs
This makes deleting frontend stacks easier as the bucket doesn't need to be emptied and deleted

### DIFF
--- a/.github/workflows/build-frontend-image.yml
+++ b/.github/workflows/build-frontend-image.yml
@@ -29,13 +29,13 @@ jobs:
 
     steps:
       - name: Push Docker image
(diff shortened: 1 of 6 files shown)


abdul-gds (approved):
looks fine

PR #385, opened 2023-08-24
SSE-2784 Service name update

rafal-surowiec-co:


rafal-surowiec-co:
> In some places you have UpdateService and others UpdateServiceName, e.g. the lambda name. I advise using ServiceName to avoid confusion - is this a service being updated or a service name?

This is Service update but allows only service name to be updated. Functions are called UpdateService and parameter type is ServiceNameUpdates. Service name is the only parameter that is updated in all service records. In future if we want to be able to modify other service parameters then backend function may work in a different way.

MartinH-DI:
Looks good, thanks

MartinH-DI (approved):

PR #419, opened 2023-09-21
Add the Dynatrace one agent to the frontend container

melvabout:
- Github workflows
    - In order to log into the dynatrace repo a PaaS token will be required and a GitHub secret, DYNATRACE_PAAS_TOKEN  creating. Information on creation of the token is here https://github.com/govuk-one-login/observability-infrastructure/tree/main/fargate#building-the-container.
- gitignore for vscode
- Cloudformation Templates
    - Giving the deployment IAM role permission to pull the secrets from di-observability-prod secrets manager
    - Pulling the values needed from the secrets
- Dockerfile change to pull the One Agent from Dynatrace container registry

Notes
- bash script needs to receive the PaaS token in someway and then authenticate against the registry. The line is there but commented out.
- Doc and what we are trying to do https://github.com/govuk-one-login/observability-infrastructure/tree/main/fargate#building-the-container
- Jira story https://govukverify.atlassian.net/browse/OBS-31

### DIFF
--- a/.github/workflows/build-frontend-image.yml
+++ b/.github/workflows/build-frontend-image.yml
@@ -33,7 +33,8 @@ jobs:
         with:
           registry: khw46367.live.dynatrace.com
           username: khw46367
-          password: ${{ secrets.DYNATRACE_PAAS_TOKEN }}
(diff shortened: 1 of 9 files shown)


melvabout:
new PR for branch with no slash in the name

PR #437, opened 2023-09-28
SSE-2858 respond to unverified change of phone number upon sign-in

MartinH-DI:
When a contact phone number is changed but not verified, the sign-in path assumes it is invalid and requests a new number and verification.


DStumpGDS:
Closing Ticket without Merging as now superseded by Branch SSE-2858-Revised (PR #452)

PR #447, opened 2023-10-10
SSE-2877 Add WAF to Cognito

rafal-surowiec-co:
…hough as there is an issue that Martin is working on.


MartinH-DI (approved):
Looking good

keith-hart (changes requested):
Dont lines 104-110 need to be uncommented now the deployment policies have been fixed

MartinH-DI (approved):

PR #468, opened 2023-10-17
SSE-2874 - password changed confirmation

MartinH-DI:
SSE-2874 - Reimplement password changed confirmation to use smtp nodemailer.


MartinH-DI:
Closing as now redundant

PR #469, opened 2023-10-18
Remove Condition from TxMA and link out own accounts to SQS Queue

JamesCarterDI:


MartinH-DI (dismissed):
Looks ok

keith-hart (changes requested):
Still seem to have the problem in https://preview-sse-2464-txmaeventlo.admin.development.sign-in.service.gov.uk

JamesCarterDI:
> Still seem to have the problem in https://preview-sse-2464-txmaeventlo.admin.development.sign-in.service.gov.uk

It was an error on my end not realising that the stacks created in CloudFormation by GitHub Actions have a prefix after 'frontend', which our normal stacks don't have. So it was working in integration, and production because we have the normal naming scheme of 'self-service-frontend' and it was able to find the SQS Queue from that.

keith-hart (approved):

PR #484, opened 2023-10-23
SSE-2949 Update allowed email domains list for production

BorisP-GDS:
Update allowed email domains list for production


MartinH-DI (dismissed):
We should come up with a better way to manage this list in the future. Approving.

MartinH-DI (approved):
Looks ok

PR #487, opened 2023-10-25
SSE-2956 improve robustness of error handling

MartinH-DI:
SSE-2956 improve robustness of error handling where a session may not exist or the remote logging service is unavailable


DStumpGDS (dismissed):
Only change I would have made is to make the getting of the Cognito ID a common private Method to save putting the same code in twice but other then that seems a good catch.

DStumpGDS (dismissed):

DStumpGDS (approved):

PR #498, opened 2023-10-30
Changed TxMA event message format

keith-hart:
Rename TxMA Events


keith-hart:
Raise as the acceptance tests were flaky on James' branch

PR #513, opened 2023-11-02
SSE-2826 removal of PasS for publish

keith-hart:


keith-hart:
Need to delete this as it wont recreate the dynamo db with the backup in place and you cant delete the backup through the AWS Console ... insufficient priviledges

PR #519, opened 2023-11-07
sse 2917 update acceptance tests

raghu-kora:
updated test setup when we enabled the browser 
Also in the current version command execution is not getting terminated after a successful run of the tests


BorisP-GDS (changes requested):
5 commits need to be squashed into one. Also, commit message must contain the story ID.

PR #525, opened 2023-11-13
SSE-2750 add allow rules for google apis

abdul-gds:
# Code Changes

- add allow domains
- add allow rules

# Why

- access to google api


MartinH-DI (approved):
I have tested this and it works. Approved.

PR #527, opened 2023-11-13
SSE-2997 - Add multi service functionality frontend

BorisP-GDS:
Add multi service functionality frontend


keith-hart:
I have checked this out and run it.  

The button is hidden behind the feature flag and adds the new service as expected.

keith-hart (approved):

PR #530, opened 2023-11-16
Sse 2779 kh

keith-hart:
This is a test to check SonarCloud


keith-hart:
Was looking at why Sonarcloud was failing ... I have opened another ticket 3019 for this as the problem is not on this branch

PR #551, opened 2023-12-07
SSE-2898 Fixing code coverage report

keith-hart:


MartinH-DI (approved):
Looks good: code coverage looks like it is more realistic.

PR #573, opened 2024-01-08
SSE-2995:Stubs required for covering the scenarios related to Expired Security Digit

DStumpGDS:
Currently, we do not have tests for covering the scenarios 

When a user tries to register with an expired security digit i.e., received in an email 

When a user tries to sign in with an expired security digit i.e., received on mobile 

 

Relevant stubs to be implemented for covering these scenarios

### DIFF
--- a/acceptance-tests/features/sign-in.feature
+++ b/acceptance-tests/features/sign-in.feature
@@ -42,6 +42,27 @@ Feature: Users can sign in to the self-service experience
       And they see the toggle link "Show" on the field "password"
       Then they can not see the content in the password field
 
+  Rule: The user tries to submit SMS security code
+    Background:
+      When they submit the email "registered@test.gov.uk"
+      And they submit a valid password
+      Then they should be redirected to the "/sign-in/enter-text-code" page
+
+    Scenario Outline: User submits a security code with invalid format
+      When they submit the value of security code "<code>" that <condition>
+      Then they should be redirected to the "/sign-in/enter-text-code" page
+      And the error message "<error_message>" must be displayed for the security code field
+      And they should see the text "We sent a code to:"
+
+      Examples:
+        | condition                   | code    | error_message |
+        | is empty                    |         | Enter the 6 digit security code |
+        | has more than 6 digits      | 1234567 | Enter the security code using only 6 digits |
+        | has fewer than 6 digits     | 12345   | Enter the security code using only 6 digits |
+        | contains letters            | 12345A  | Your security code should only include numbers |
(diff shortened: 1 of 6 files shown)


acceptance-tests/features/sign-in.feature
```diff
@@ -42,6 +42,47 @@ Feature: Users can sign in to the self-service experience
       And they see the toggle link "Show" on the field "password"
       Then they can not see the content in the password field
 
+  Rule: The user tries to sign in with an expired or invalid SMS Code
```
BorisP-GDS, line 45: Rename to : The user tries to submit SMS security code

BorisP-GDS (changes requested):

acceptance-tests/features/sign-in.feature
DStumpGDS: Thanks Boris, change made as requested.

DStumpGDS:
Thanks for this suggestion Boris - Now implemented.

acceptance-tests/support/steps/input-steps.ts
```diff
@@ -62,3 +62,9 @@ Then("the value of the text field {} should be {string}", async function (this:
     );
     assert.equal(elementAttributeValue, attributeValueToCheck);
 });
+
```
BorisP-GDS, line 65: To avoid linter errors insert the following line here : 
```
// eslint-disable-next-line
```
In this case disabling linter makes sense, because we need the parameter for the function to work as intended (to insert the condition in the step definition), but we do not need to use it inside the function.

acceptance-tests/support/steps/input-steps.ts
```diff
@@ -62,3 +62,9 @@ Then("the value of the text field {} should be {string}", async function (this:
...
+When("they submit the value of {} {string} that {}", async function (fieldName, value, condition) {
+    console.log(JSON.stringify(condition));
```
BorisP-GDS, line 67: Remove this line (the console.log ), we do not need the values logged out, logging out is used for debugging.

acceptance-tests/support/steps/input-steps.ts
DStumpGDS: > // eslint-disable-next-line

Boris, are you sure this will resolve the ESLINT Issue which was that Condition is being declared but not used, hence my Console.log?

BorisP-GDS: As you can see, there are no lint issues in the latest build.

BorisP-GDS (approved):
LGTM

PR #584, opened 2024-01-16
SSE-2872:Logging visibility to be updated when service name not regis…

DStumpGDS:
Scenario: when service registration failed while creating an account 

relevant user cannot able to access the service and the user is not getting an option to either to create a service(ideal way) or not getting access to service which is displaying while logging in to application
stepFunction: RegisterAuthClient

### DIFF
--- a/express/src/controllers/register.ts
+++ b/express/src/controllers/register.ts
@@ -333,9 +333,20 @@ export const processAddServiceForm: RequestHandler = async (req, res, next) => {
 
     req.session.serviceName = req.body.serviceName;
 
-    const generatedClient = await s4.generateClient(service, nonNull(req.session.authenticationResult));
-    const body = JSON.parse(generatedClient.data.output).body;
-    const serviceId = JSON.parse(body).pk;
+    let serviceId = "";
+    let body;
+
+    try {
+        console.info("Generating Client to Service:" + service.serviceName);
+        const generatedClient = await s4.generateClient(service, nonNull(req.session.authenticationResult));
+        body = JSON.parse(generatedClient.data.output).body;
+        serviceId = JSON.parse(body).pk;
+    } catch (error) {
+        console.error("Unable to Register Client to Service - Service Items removed");
+        console.error(error);
+        await s4.deleteServiceEntries(uuid);
+        return res.render("there-is-a-problem.njk");
+    }
 
     req.session.serviceId = serviceId;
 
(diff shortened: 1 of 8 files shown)


express/src/controllers/register.ts
```diff
@@ -394,3 +405,11 @@ export const resumeUserJourneyAfterPassword: RequestHandler = async (req, res) =
     // This is to allow for the fact that they may have entered an incorrect number.
     return res.redirect("enter-phone-number");
 };
+
+export const rolbackServiceEntriesForClient: RequestHandler = async (req, res) => {
```
MartinH-DI, line 409: rolbackServiceEntriesForClient -> rollbackServiceEntriesForClient ??

DStumpGDS:
Good spot Martin! - Method name now changed to correct one.

express/src/controllers/register.ts
DStumpGDS: Good spot Martin! - Method now changed.

MartinH-DI (dismissed):
Looks good

express/src/controllers/register.ts
```diff
@@ -333,9 +333,20 @@ export const processAddServiceForm: RequestHandler = async (req, res, next) => {
...
+    try {
+        console.info("Generating Client to Service:" + service.serviceName);
+        const generatedClient = await s4.generateClient(service, nonNull(req.session.authenticationResult));
+        body = JSON.parse(generatedClient.data.output).body;
+        serviceId = JSON.parse(body).pk;
+    } catch (error) {
+        console.error("Unable to Register Client to Service - Service Items removed");
+        console.error(error);
+        await s4.deleteServiceEntries(uuid);
+        return res.render("there-is-a-problem.njk");
```
rafal-surowiec-co, line 348: return res.render("rollback-service-entries.njk");

rafal-surowiec-co (changes requested):

express/src/controllers/register.ts
DStumpGDS: Hi Rafal, I think the actual solution is not what you suggested but to remove the rollbackServiceEntriesForClient Request Handler and Routes associated with it - I think this was originally put in to do the Logic of removing the Entries but resulted in just redirecting to the Problem Page which I now do in the Catch Section anyway (the Handler is effectively redundant).

DStumpGDS: Rafal has agreed to removal of redundant Request Handler and associated Route.

rafal-surowiec-co (approved):

PR #587, opened 2024-01-23
SSE-2994:Created PII Scanner Script which run as part of Acceptance Tests and also available as Command Line Option

DStumpGDS:
Tests to be implemented for Log Entries that cover part of Personal Identity Information


MartinH-DI (approved):
Looks good

acceptance-tests/package.json
```diff
@@ -8,6 +8,6 @@
   },
   "devDependencies": {
     "@cucumber/cucumber": "^9.1.0",
-    "puppeteer": "^20.7.2"
+    "puppeteer": "^20.3.0"
```
rafal-surowiec-co, line 11: Any reason we are downgrading this ?

DStumpGDS: I can only assume this was done as part of Andrew's fixes so suggest you speak to him about the reasons - He was looking into why the Build was failing and got it working.

If you wish it to be re-instated please let me know and I will be happy to comply.

PR #591, opened 2024-01-26
SSE-3110: Explicit specification of log groups with Data Protection Policy

kalpaitch:
This PR aims to explicitly declare the LogGroups used by the core services in the stack and apply Data Protection Policies to each to redact PII, goes hand in hand with https://github.com/govuk-one-login/onboarding-product-page/pull/276.

This includes separation of log groups **so that admin-tool and product pages don't share the same log groups** (and can therefore manage their own policies). This essentially means adding an additional application separator to the LogGroupPrefix, which changes from `/self-service` to `/self-service/[ product-pages | admin-tool ]`.


abdul-gds (approved):
looks good

abdul-gds (approved):
looks good

PR #600, opened 2024-01-31
SSE-3110: Updated log permissions for cognito build

kalpaitch:
I've deployed these changes to the deployment-config.

Does this also require a secure pipeline update? Or will the changes all be picked up by the next run?


abdul-gds:
@kalpaitch these are for dev env only
and we need to apply manually by [script](https://github.com/govuk-one-login/onboarding-self-service-experience/blob/91e3bbec0c28fb3bdcb758ad79193bc5b3db05e6/infrastructure/ci/development/apply-deployment-config.sh)

abdul-gds (approved):
looks good

PR #607, opened 2024-02-05
SSE-3150: Removed LogGroupPrefix as input parameter

kalpaitch:
Taking into consideration that [Secure Pipelines does not publish custom input parameters](https://govukverify.atlassian.net/wiki/spaces/PLAT/pages/3193831452/Stack+parameters+provided+by+the+pipeline), it is best to vary the [CloudFormation stack configuration](https://govukverify.atlassian.net/wiki/spaces/PLAT/pages/3071705136/How+to+create+a+configuration+pipeline#Step-2%3A-Use-CloudFormation-mappings-to-vary-the-parameter-values-by-environment) based on `Environment` instead.

Because the value of `LogGroupPrefix` is already set (to the default value) for established environments it is necessary to remove this completely (because the default value can not now be changed).

Setting some rules for `DeploymentName`, so that this only varies from the default value for ephemeral environments.

### DIFF
--- a/.github/workflows/deploy-branch.yml
+++ b/.github/workflows/deploy-branch.yml
@@ -34,7 +34,6 @@ jobs:
       template: infrastructure/frontend/frontend.template.yml
       parameters: |
         ImageURI=${{ needs.push-frontend-image.outputs.image-uri }}
-        LogGroupPrefix=/self-service/admin-tool/preview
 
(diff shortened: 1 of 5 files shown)


abdul-gds (dismissed):
looks good

abdul-gds (approved):
looks good

PR #652, opened 2024-03-05
SSE-3229: Passing through new fields `back_channel_logout_uri` and `sector_identifier_uri`

Ryan-Andrews99:
### This PR Adds
- Stores  the fields  `back_channel_logout_uri` and `sector_identifier_uri` from the Client Registry API Response in the dynamo table
- These fields are passed to the server when fetched from the dynamo table
- Unit tests for the putServiceClientHandler
- A setupFile for the API Jest config which sets up shared environment variables 
- The updating of these new fields is already handled when calling our `update-client` endpoint. 
> [!NOTE]
>  The Client Registry API does some validation when trying to update these values so some client side validation before submitting an update would be useful to avoid 400 Errors (https://govukverify.atlassian.net/wiki/spaces/DFA/pages/3931734342/Client+registry+-+Fields)

### This PR Modifies:
- Existing Unit tests in `backend/api/tests/dynamodb-client.test.ts` are modified to use the global setup env var
- The type in `putServiceClientHandler` is updated to remove the need for an explicit `any`: 
> - The event body is typed as a clientRegistryRegistrationResponse which includes all the fields from the previous lambda in the state machine
> - I have removed the typing of the lambda invoke event as APIGatewayProxyEvent. This is because this event type corresponds to a lambda function invoked by an APIGateway. This lambda is not invoked by an API GW instead it is invoked with a defined JSON Payload from the state machine

### This PR Removes: 
- N/A


BorisP-GDS (changes requested):
Could you, please, squash 10 commits into one, the commit message should start with your story number (in this case SSE-3229 ) and short text to describe the work or if it suits you can use the name of the story.

rafal-surowiec-co (approved):

BorisP-GDS (approved):
LGTM

PR #662, opened 2024-03-08
SSE-2966:Added Fixed OTP Processing Capability for Test Accounts to bypass Email and SMS Codes

DStumpGDS:
An agreed solution using an Environment Parameter USE-STUB-OTP which when set to TRUE will allow Fixed OTP Processing or if FALSE then any such support is blocked. The Fixed OTP Parameters will be defined in the Secret FIXED-OTP-CREDENTIALS.

A full description (E&OE) of this process can be found in the Confluence Document [Re: Technical Strategy for supporting Fixed OTP in End to End Testing | Comment](https://govukverify.atlassian.net/wiki/spaces/DFA/pages/3992649857/Technical+Strategy+for+supporting+Fixed+OTP+in+End+to+End+Testing?focusedCommentId=3994746950)


rafal-surowiec-co (commented):
David, can you provide unit test as well ? We need to start improving coverage.

DStumpGDS:
@Rafal Surowiec ***@***.***> - These are
being developed now as part of SSE-2968.

express/src/lib/fixedOTPSupport.ts
MartinH-DI, line 1: There is quite a bit of info being logged out in these functions. Do we want to leave that in?

DStumpGDS: Thanks for spotting that Martin - All but the 'Breadcrumb' Loggin has now been removed.

MartinH-DI (approved):
Looking good

PR #664, opened 2024-03-11
SSE-3232 Extend User Attributes section of Client page

MartinH-DI:
SSE-3232 Extend User Attributes section of Client page. Add processing of the Identity Verification toggle, and the conditionally visible Claims field.


backend/api/src/handlers/dynamodb/put-service-client.ts
```diff
@@ -34,8 +34,8 @@ export type putServiceHandlerInvokeEvent = {
     body: string;
 };
 
-export const putServiceClientHandler = async (event: putServiceHandlerInvokeEvent): Promise<APIGatewayProxyResult> => {
-    const payload: clientRegistryRegistrationResponse = JSON.parse(event.body);
+export const putServiceClientHandler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
```
Ryan-Andrews99, line 37: I added a different type on `event` previously as this lambda was not invoked by API gateway but instead invoked by a step function with a defined JSON Payload, is there a reason this has changed back?

backend/api/src/handlers/dynamodb/put-service-client.ts
```diff
@@ -34,8 +34,8 @@ export type putServiceHandlerInvokeEvent = {
...
+    const payload = JSON.parse(event.body as string);
```
Ryan-Andrews99, line 38: Same as above but I previously did some type annotation to give a bit more type safety to what the body of the event looked like, not sure why it's been removed

backend/api/tests/handlers/dynamodb/put-service-client.test.ts
```diff
@@ -62,15 +68,24 @@ const expectedDynamoRecord = {
...
+        process.env = originalEnv;
     });
 
     it("calls the dynamo client with a put command with the expected values and returns a 200 with the expected response body", async () => {
         const putItemSpy = jest.spyOn(DynamoDbClient.prototype, "put").mockResolvedValue({
             $metadata: {}
         });
 
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
Ryan-Andrews99, line 87: Another `@ts-ignore` comment, I know the concept of invoking the lambda with a status code is odd, but that is the shape of the incoming data if you look at the step function in the console

backend/api/tests/handlers/dynamodb/put-service-client.test.ts
```diff
@@ -85,8 +100,9 @@ describe("putServiceClientHandler tests", () => {
         const error = "SomeAwsError";
         const putItemSpy = jest.spyOn(DynamoDbClient.prototype, "put").mockRejectedValue(error);
 
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
Ryan-Andrews99, line 104: Same as above

backend/api/tests/handlers/dynamodb/update-service-client.test.ts
```diff
@@ -0,0 +1,77 @@
...
+            body: '{"$metadata":{}}'
+        });
+    });
+
+    it("calls the dynamo client with a update command with the expected values and returns a 500 when the dynamo client throws an error", async () => {
+        const error = "SomeAwsError";
+        const updateItemSpy = jest.spyOn(DynamoDbClient.prototype, "updateClient").mockRejectedValue(error);
+
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
Ryan-Andrews99, line 61: `@ts-ignore` again

backend/api/tests/handlers/dynamodb/update-service-client.test.ts
Ryan-Andrews99, line 1: Very happy to see our test coverage going up!

backend/api/src/handlers/dynamodb/put-service-client.ts
MartinH-DI: Thanks. Yes: this makes sense and I've worked through cleaning these up.

MartinH-DI: Thanks. Yes: this makes sense and I've worked through cleaning these up.

backend/api/tests/handlers/dynamodb/put-service-client.test.ts
MartinH-DI: Thanks. Yes: this makes sense and I've worked through cleaning these up.

backend/api/tests/handlers/dynamodb/put-service-client.test.ts
MartinH-DI: Thanks. Yes: this makes sense and I've worked through cleaning these up.

backend/api/tests/handlers/dynamodb/update-service-client.test.ts
MartinH-DI: Thanks. Yes: this makes sense and I've worked through cleaning these up.

backend/api/src/handlers/dynamodb/get-service-clients.ts
```diff
@@ -1,17 +1,19 @@
-import {APIGatewayProxyEvent, APIGatewayProxyResult} from "aws-lambda";
 import DynamoDbClient from "../../dynamodb-client";
 
 const client = new DynamoDbClient();
 
-export const getServiceClientsHandler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
-    const serviceId = event.pathParameters?.serviceId;
-    if (serviceId === undefined) {
+export type getServiceHandlerInvokeEvent = {
```
Ryan-Andrews99, line 5: Sorry if this seems a but back and forth-y, but I do think this handler **is** invoked by an API gateway at the endpoint `/get-services/{userId}`, so I think we should leave this type as is. It is bit confusing because there are a lot of lambdas in our backend and it's hard to know what is state machine invoked and what is API Gateway invoked. In this case you can find the handler in the `api.template.yml` with this section: 
```
Events:
        Api:
          Type: Api
          Properties:
            RestApiId: !Ref API
            Path: /get-services/{userId}
            Method: GET
```

MartinH-DI: Thanks: it is a fair point. Reverted this and fixed the unit test to accommodate .

Ryan-Andrews99 (approved):

PR #680, opened 2024-03-21
SSE-3278: Template value not needed after a sam build.

kalpaitch:


.github/workflows/deploy-branch.yml
```diff
@@ -47,7 +47,6 @@ jobs:
     with:
       name: ${{ needs.build-cognito.outputs.component }}
       artifact: ${{ needs.build-cognito.outputs.artifact-name }}
-      template: backend/cognito/cognito.template.yml
```
abdul-gds, line 50: why deleting input for template?

PR #682, opened 2024-03-22
SSE-2968:Added Example Tests which use the new OTP Funcitonality

DStumpGDS:
This Commit contains two example Cucumber Tests which use the new OTP Functionality to create an Account and then Sign-Into it.

These tests are not designed for 'Production' Testing but can be used as Templates to produce more robust and targeted Tests or to explore  the fixedOTP Functionality.

One major caveat is the Tests are hostage to the idiosyncrasies of the Backend Cognito Interaction and therefore can fail quite often but *** THIS IS NOT *** a problem with the Tests OR the OTP Functionality but is a know issue with the application and something which can occur quite often even when using the application manually and this needs to borne in mind when running these Tests as it can be quite frustrating!!

### DIFF
--- a/.github/workflows/run-exampleotp-tests.yml
+++ b/.github/workflows/run-exampleotp-tests.yml
@@ -1,0 +1,67 @@
+name: exampleotp tests
+
+on:
+  workflow_call:
+    inputs:
+      environment: { required: true, type: string }
+
+permissions: { }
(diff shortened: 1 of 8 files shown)


DStumpGDS:
Removed due to Credential Violation which needs to be resolved.

DStumpGDS:
Removed due to Credential Violation which needs to be resolved.

PR #698, opened 2024-04-03
SSE-2968:Tests to check the behaviour

DStumpGDS:
Allow automatic testing by providing Fixed OTP Support Functionality which will allow 'known' values to be inserted automatically so no Email or Telephone SMS Intervention will be required to fulfil Registration or Sign-in. 

All Entry Fields can now be used with actual values or Pseudo ID's in which case the values would be pulled from a Secret (FIXED-OTP-CREDENTIALS).

Two Example Tests using this have been provided - Please note the Tests when run can be 'flaky' but this is due to external issues around the general operation of the Server Back-End as well as more generic Network Timing Issues.


ui-automation-tests/exampleotp-features/register.feature
```diff
@@ -0,0 +1,25 @@
+Feature: Users can sign up to the self-service experience using real data.
+
+  Background:
+    Given the user is on the "/register" page
+
+  Rule: Register Account
+    Scenario: The user registers a new account
+      When they submit the email "~Registered"
```
rafal-surowiec-co, line 8: If feature is enabled in prod by mistake and they provide '~Registered' then they will be able to create account without any verification.

DStumpGDS: Yes you are quite correct but that is also true if we simply had fixed Email Addresses and Fixed OTP Codes etc.. 

I am a bit confused since we have discussed all this previously and agreed a solution which you are now concerned about. 

Any fixed OTP System will, if allowed to run in Production, will allow free reign that is why we have the Flag USE_STUB_OTP to prevent this and, to my mind, is no different to the approach adopted by other teams whereby they use similar flags; Please can I refer you to other teams approach where they have a TEST_CLIENTS_ENABLED Flag (so same thing but different name) which is only set in the Environments where they wish the functionality to work (please see https://govukverify.atlassian.net/wiki/spaces/LO/pages/3503128761/Fixed+One+Time+Password+OTP+for+test+users)

MartinH-DI (approved):
This all looks good to me

rafal-surowiec-co (approved):

PR #708, opened 2024-04-22
BAU: `sign-in` and `accounts` controller unit tests

Ryan-Andrews99:
### This PR Adds:
- Unit tests where they are missing for the `sign-in.ts` and `account.ts` controllers
- Two small code changes in the `changePassword` controller and `forgotPassword` controller. These add an errorMessages property on an options object to resolve a runtime TypeError

### This PR Modifies:
- Some of the existing Unit tests as I have moved the mock S4 instance into the main request mock instead of creating one each time a mock request is created 

### This PR Removes:
- N/A


MartinH-DI (approved):
This looks good.

PR #730, opened 2024-05-10
SSE-3279:Added specific Expired Codes for Email and OTP to provide known Testing Parameters

DStumpGDS:
# Onboarding Feature Deployment

> [!WARNING]
> Pull requests merged to `main` will be released to production, please ensure the checklist below is complete

Before any work can be merged to main in must meet the definition of done and be ready to deploy. While many of these tasks will be automated, the reviewers must take the responsibility of confirming the checklist below has been completed before this ticket can be merged.

## Checklist

-   [ X] this pull request meets the acceptance criteria of the ticket

-   [ X] this branch is up-to-date with the main branch

    `git fetch --all && git rebase origin/main`

-   [ X] these changes are backwards compatible (no breaking changes)

    -   all methods signatures and return values are the same
    -   any replaced methods are marked as `@deprecated`

-   [ ] tests have been written to cover any new or updated functionality

-   [ ] new configuration parameters have been deployed to all environments, see [configuration management](https://govukverify.atlassian.net/l/cp/N7q3Vh3r).

-   [ ] all external infrastructure dependencies have been updated in all environments

## Changes

[ _please list the changes this pull request is making_ ]


MartinH-DI (approved):
This looks good

rafal-surowiec-co (commented):
David, can you add unit tests ? Coverage of file modified is very low
```
fixedOTP.ts                            |   29.11 |     4.76 |   14.28 |   29.48 | 21-88,92-99,105-106,110,119,127,134-135,139-140,144-145,149-150,154-155,159-160
```

MartinH-DI (dismissed):
Looks good

express/tests/lib/fixedOTP.test.ts
```diff
@@ -0,0 +1,406 @@
...
+        } catch (error) {
+            checkResult = false;
+
+            if (error instanceof CodeMismatchException) {
+                checkResult = expectException;
+            }
+        }
+
+        expect(checkResult).toBe(true);
+    });
```
Ryan-Andrews99, line 272: I am not too sure what these tests actually do, not sure if they need to be separate `it` blocks or not? I've commented out all of the calls to `respondToMFAChallengeForFixedOTPCredential` and they still seem to pass

express/tests/lib/fixedOTP.test.ts
```diff
@@ -0,0 +1,406 @@
...
+            verifyMobileUsingOTPCode("testuser1@unit-testing.com", "999999");
+        } catch (error) {
+            checkResult = false;
+
+            if (error instanceof CodeMismatchException) {
+                checkResult = expectException;
+            }
+        }
+
+        expect(checkResult).toBe(true);
```
Ryan-Andrews99: Same here as below, these tests pass regardless of whether `verifyMobileUsingOTPCode` is called

express/tests/lib/fixedOTP.test.ts
Ryan-Andrews99, line 1: Really happy to see more tests being added! Most of these are great but some of them don't seem to directly test the code in `fixedOTP` and look like they might always pass. Not exactly sure how we can test these more directly but happy to pick this up with you if that would be helpful

DStumpGDS: I have done unit Tests for all the Exported Functions which will then test all the Internal Functions also.

I have also created 3 different Test 'Scenarios' with the results for each Test being amended to reflect what is expected to be returned and these are:

1. OTP Flag (USE_STUB_OTP) not set
2. OTP Flag set but Blank Credentials (Secret)
3. OTP Flag set and Credentials

express/tests/lib/fixedOTP.test.ts
DStumpGDS: Yes because the Exception is only raised when the Method is called with an Invalid Value so if you don't call anything the Tests will pass but that is not a Valid Unit Test!

express/tests/lib/fixedOTP.test.ts
DStumpGDS: Yes because the Exception is only raised when the Method is called with an Invalid Value so if you don't call anything the Tests will pass but that is not a Valid Unit Test!

Ryan-Andrews99: Ah I get what you mean. There is a test matcher here for error throwing https://jestjs.io/docs/expect#tothrowerror, it does look a bit weird but that's just javascript weirdness 😄 . You could do something like this to test the error throwing more directly:

```
expect(() =>  verifyMobileUsingOTPCode("testuser1@unit-testing.com", "123456")).toThrow(CodeMismatchException)
```

This way the test fails if there is no exception thrown

DStumpGDS:
Thanks Ryan, that looks really useful.

Only problem is not all the calls here are designed to throw an Error (as
you can discern from the code) so with your solution how would I catch the
Error (and any other possible error) being thrown when it shouldn't be?

express/tests/lib/fixedOTP.test.ts
DStumpGDS: New tests added as suggested.

Ryan-Andrews99 (approved):

rafal-surowiec-co (approved):

PR #778, opened 2024-06-06
SSE-3275: Update dependabot config for NPM workspaces

ethanmills:
As described in https://github.com/dependabot/dependabot-core/issues/5226

The previous dependabot config dig not update the package-lock file


ethanmills:
Because they're all workspaces in the topmost package.json, I believe this should still update all of the modules. Could be wrong, but that's my understanding and the intention!

PR #903, opened 2024-09-12
Ato 1022 fix update service access

isaac-GDS:
Add checks to updateService to allow access.

Replicating the service and user check in the ListClient to the updateService. As some forms dont call the ListClient this is another layer of protection in assuring the user has access to the client update.

### DIFF
--- a/backend/api/api.template.yml
+++ b/backend/api/api.template.yml
@@ -946,6 +946,9 @@ Resources:
       Policies:
         - AWSLambdaVPCAccessExecutionRole
         - AWSXrayWriteOnlyAccess
+        - DynamoDBReadPolicy:
+            TableName:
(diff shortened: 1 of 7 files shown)


backend/api/api.template.yml
```diff
@@ -946,6 +946,9 @@ Resources:
       Policies:
         - AWSLambdaVPCAccessExecutionRole
         - AWSXrayWriteOnlyAccess
+        - DynamoDBReadPolicy:
+            TableName:
+              Fn::ImportValue: !Sub ${DeploymentName}-DynamoDB-DataTableName
```
CarlyG55, line 951: just to confirm, you've tested the changes that need these permissions changes?

Ryan-Andrews99: Yep we ran through a journey of creating an account + service and then updating the new service to check this worked

CarlyG55 (approved):

PR #970, opened 2024-12-16
Ato 1237/update consumer tests

isaac-GDS:
# Onboarding Feature Deployment

Consumer test for POST and PUT request for client registry

Updated test to call handler instead of mocking out.
Added type for response object. (Maybe assert it in handler later)
Remove initial setup (verification)

## How to review
 - make sure no changes to production code
 - check we are asserting the right checks for the data in the pact contract


backend/api/tests/contract-tests/tests/testClientRegistryContract.spec.ts
```diff
@@ -73,61 +23,121 @@ describe("ClientRegistryProvider", () => {
...
+        const client_config = {
+            client_name: postEvent.service.serviceName,
+            public_key: public_key,
+            redirect_uris: ["http://localhost/"],
+            contacts: [postEvent.contactEmail],
+            scopes: ["openid", "email", "phone"],
+            subject_type: "pairwise",
+            service_type: "MANDATORY",
+            sector_identifier_uri: "http://gov.uk",
+            client_locs: ["P2"]
```
isaac-GDS, line 43: client locs are not returned in the response body - is this okay ?

backend/api/src/handlers/types/client-registry-response.ts
```diff
@@ -0,0 +1,23 @@
...
+  back_channel_logout_uri: null | string;
+  token_endpoint_auth_method: string;
+  response_type: string;
+  jar_validation_required: boolean;
+  claims: string[];
+  client_type: string;
+  id_token_signing_algorithm: "ES256" | "RS256";
+  jwks_uri: null | string;
+  channel: string;
+  max_age_enabled: boolean;
```
isaac-GDS, line 22: These two are new - dont know if this expected elsewhere in the code

backend/api/tests/contract-tests/tests/testClientRegistryContract.spec.ts
```diff
@@ -73,61 +23,121 @@ describe("ClientRegistryProvider", () => {
...
                     }
                 });
 
             await provider.executeTest(async mockProvider => {
-                const [results] = await Promise.all([postRequest(mockProvider.url, EXPECTED_BODY)]);
-                console.log("RESULTS: " + JSON.stringify(results));
-                expect(results).toContain(EXPECTED_BODY);
+                process.env.AUTH_REGISTRATION_BASE_URL = mockProvider.url;
+                const results = await registerClientHandler(postEvent, mockLambdaContext);
+                expect(JSON.parse(results.body)).toMatchObject(expected_returned_values);
```
isaac-GDS, line 72: What is more important to check in here. That the response body has the values that is sent or it is a client registry object.

backend/api/tests/contract-tests/tests/testClientRegistryContract.spec.ts
```diff
@@ -73,61 +23,121 @@ describe("ClientRegistryProvider", () => {
...
                     }
                 });
 
             await provider.executeTest(async mockProvider => {
-                const [results] = await Promise.all([putRequest(mockProvider.url, EXPECTED_BODY)]);
-                console.log("RESULTS: " + JSON.stringify(JSON.stringify(results)));
-                expect(results).toContain(EXPECTED_BODY);
+                process.env.AUTH_REGISTRATION_BASE_URL = mockProvider.url;
+                const results = await updateClientInRegistryHandler(putEvent, mockLambdaContext);
+                expect(JSON.parse(results.body)[0]).toMatchObject(expected_response);
```
isaac-GDS, line 140: This is because in the handler it spreads `{ "0": { ....}}` Is this in purpose, it should probably spread the object inside "0".

backend/api/src/handlers/types/client-registry-response.ts
Ryan-Andrews99: The current code is quite permissive - these will have been stored if they are included in the response. But good we have this type, we should try and use it in most places in the codebase!

backend/api/tests/contract-tests/package.json
```diff
@@ -8,8 +8,8 @@
     "buildjs": "cp -r assets/javascripts dist/assets/",
     "build": "npm run buildts && npm run buildjs",
     "publish": "node dist/provider/publish.js",
-    "verify": "node dist/provider/verifier.js",
-    "test": "npm run publish && jest --detectOpenHandles"
+    "test": "npm run publish && jest --detectOpenHandles",
+    "test:pact": "jest --detectOpenHandles"
```
Ryan-Andrews99, line 12: Think we can bin this `--detectOpenHandles` flag, [it's probably not useful here](https://jestjs.io/docs/cli#--detectopenhandles)

backend/api/tests/contract-tests/tests/testClientRegistryContract.spec.ts
Ryan-Andrews99: Ah yeah sort of remember finding this -  https://govukverify.atlassian.net/browse/ATO-1040. We can fix this as part of the above ticket

backend/api/tests/contract-tests/tests/testClientRegistryContract.spec.ts
Ryan-Andrews99: Hmm I think we should match the same level of validation currently in the code - it's quite permissive right now. If we start asserting the values a bit more - these tests can be updated

backend/api/tests/contract-tests/tests/testClientRegistryContract.spec.ts
Ryan-Andrews99: Hmm not sure I understand this one?

Ryan-Andrews99: Hmm not sure I understand this one?

backend/api/tests/contract-tests/tests/testClientRegistryContract.spec.ts
Ryan-Andrews99: The current code is quite permissive, so I think we should match that level of assertion. If we do more validation in the handler we should update these tests!

backend/api/tests/contract-tests/tests/testClientRegistryContract.spec.ts
Ryan-Andrews99: Yep remember coming across this a while ago: https://govukverify.atlassian.net/browse/ATO-1040 - we can fix this as part of this ticket

backend/api/tests/contract-tests/tests/testClientRegistryContract.spec.ts
isaac-GDS: We are spreading the object in the handler , but due to the object level we are just spreading a single object `{ "0": { ....}}` instead of whats inside that object. We should probably spread what inside or just set the object with a key instead of `0`

isaac-GDS: This was due to the Matcher using `eachLike` which assert on arrays. Updated the matcher so this comment is no longer applicable.

backend/api/tests/contract-tests/tests/testClientRegistryContract.spec.ts
```diff
@@ -73,62 +24,145 @@ describe("ClientRegistryProvider", () => {
...
                     }
                 });
 
             await provider.executeTest(async mockProvider => {
-                const [results] = await Promise.all([postRequest(mockProvider.url, EXPECTED_BODY)]);
-                console.log("RESULTS: " + JSON.stringify(results));
-                expect(results).toContain(EXPECTED_BODY);
+                process.env.AUTH_REGISTRATION_BASE_URL = mockProvider.url;
+                const results = await registerClientHandler(postEvent, mockLambdaContext);
+                expect(JSON.parse(results.body)).toMatchObject(expected_response);
```
isaac-GDS, line 87: This has highlighted that in lambda code when it spreads the response object it overwrites the configuration as fields are the same name. This does not affect the contract tests as we only care about the response but the code is misleading as it the whole configuration object is overwritten.

backend/api/src/handlers/types/client-registry-response.ts
Ryan-Andrews99: Hmm we don't do anything with them yet but we probably will in the future!

Ryan-Andrews99 (commented):
Looks good, some tiny comments - good commit message on: faec7c6c2f6c34d2bebae7d5a7b1771c6a2571ff

backend/api/tests/contract-tests/tests/testClientRegistryContract.spec.ts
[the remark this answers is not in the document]
isaac-GDS: Can do !

backend/api/tests/contract-tests/package.json
```diff
@@ -8,7 +8,8 @@
     "buildjs": "cp -r assets/javascripts dist/assets/",
     "build": "npm run buildts && npm run buildjs",
     "publish": "node dist/provider/publish.js",
-    "test": "npm run publish && jest --detectOpenHandles"
+    "test": "npm run publish && jest",
```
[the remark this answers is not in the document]
isaac-GDS: Good point

backend/api/tests/contract-tests/tests/testClientRegistryContract.spec.ts
isaac-GDS: Will add this in a ticket in the SSE backlog

backend/api/tests/contract-tests/package.json
```diff
@@ -9,7 +9,7 @@
     "build": "npm run buildts && npm run buildjs",
     "publish": "node dist/provider/publish.js",
     "test": "npm run publish && jest",
-    "test:pact": "jest"
```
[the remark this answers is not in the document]
Ryan-Andrews99: Wrong package.json 😅

Ryan-Andrews99 (approved):

Ryan-Andrews99 (approved):

PR #1086, opened 2025-06-30
ATO-1730/Allow RP to set landing page url

james-peacock-gds:
### Wider context of change

RPs should have a landing page url registered if they are going to use identity verification. The ability to add landing page url to client config is being [added to authentication api](https://github.com/govuk-one-login/authentication-api/pull/6715)

### What’s changed

- add landingPageUrl to the 'Manage client' settings on Manage service page, with a corresponding page for updating landingPageUrl
- add hint text to the enable identity verification page to explain the need for landing page url
- conditionally add reminder about landing page url to the success banner when identity verification enabled
- generate success banner text in the controller to avoid complex logic in the template

NB: `landingPageUrl` is the term used in the Client Registry so this term is used in the SSE code. However, in the UI other urls are referred as URI (e.g. Sector identifier URI) so to be consistent with these other settings labels, the label 'Landing Page URI' is displayed in the UI.

### Manual testing

Deployed to dev alongside [changes to authentication-api](https://github.com/govuk-one-login/authentication-api/pull/6715) and observed the following successful test cases:

**1. Create a new service, which by default has no landingPageUrl set**
Screenshot 2025-06-26 at 14 53 20

**2. Hint text on enable identity verification page**
Screenshot 2025-06-26 at 14 53 44

**3. Success banner for enabling identity verification changes when no landingPageUrl set**
Screenshot 2025-06-26 at 14 54 09

**4. Success banner for enabling identity verification changes when landingPageUrl already set**
Screenshot 2025-06-26 at 14 54 34

**5. Success banner for disabling identity verification changes**
Screenshot 2025-06-26 at 14 53 57

**6. Success banner for adding/changing landingPageUrl**
Screenshot 2025-06-26 at 14 54 22

**7. Invalid landingPageUrl entered**
Screenshot 2025-06-26 at 14 54 47

- also checked that adding/updating landingPageUrl was reflected in the client registry dynamo db table.

### Related PRs

- [changes to authentication-api](https://github.com/govuk-one-login/authentication-api/pull/6715) to accept landingPageUrl in client registration and updates

### DIFF
--- a/express/src/controllers/clients.ts
+++ b/express/src/controllers/clients.ts
@@ -38,12 +38,16 @@ export const showClient: RequestHandler = async (req, res) => {
         displayedKey = client.jwksUri;
     }
 
+    const successBannerMessage = req.session.updatedField
+        ? generateSuccessBannerMessage(req.session.updatedField, identityVerificationSupported, landingPageUrl)
+        : undefined;
+
     res.render("clients/client-details.njk", {
         clientId: authClientId,
         selfServiceClientId: selfServiceClientId,
         serviceId: serviceId,
         serviceName: serviceName,
-        updatedField: req.session.updatedField,
+        successBannerMessage: successBannerMessage,
         redirectUris: redirectUris,
         scopesRequired: client.scopes,
         ...(client.token_endpoint_auth_method === "client_secret_post"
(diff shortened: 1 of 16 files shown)


express/src/controllers/clients.ts
```diff
@@ -990,3 +995,35 @@ export const processChangePKCEEnforcedForm = async (req: Request, res: Response)
...
+
+export const processChangeLandingPageUrlForm: RequestHandler = async (req: Request, res: Response): Promise<void> => {
+    const landingPageUrl = req.body.landingPageUrl;
+    const s4: SelfServiceServicesService = req.app.get("backing-service");
+
+    await s4.updateClient(
+        nonNull(req.context.serviceId),
+        req.params.selfServiceClientId,
+        req.params.clientId,
+        {landing_page_url: landingPageUrl},
```
cearl1, line 1039: Do we need any validation here to check if the url is valid? Not sure if it will complain about this when we retrieve it from the client registry if we save it here and it's invalid.

